Issue #1, opened 2019-06-19
merge CardContent into Card component


txwkx:
Overall the motivation behind this change is to reduce the number of available components.

- add _css_ styles:
```
width: 100%;
```

- provide the following list of _props_:
- [x] max-width
- [x] margins

PR #2, opened 2019-07-02
🔀 Merge CardContent and Card together

lapostoj:
Addressing https://github.com/marshmallow-insurance/smores-react/issues/1


txwkx:
@lapostoj I've slightly adjusted the **props** (switch to optional and change type for margins), take a look

lapostoj:
@txwkx good point about having them optional!
For the margin though, having the number would prevent to use stuff like "auto", right? 😕 
About auto, should that be the default rather than 0? 🤔

txwkx:
@lapostoj true, my bad. reverting the *margin* types back to `string`

PR #48, opened 2019-09-26
Remove Yarn

alexwllms:
Remove Yarn from the project:

- Update Travis file to use NPM
- Update Readme file
- Remove yarn.lock


src/CheckBox/__tests__/__snapshots__/CheckBox.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`renders 1`] = `
 <label
-  class="sc-htpNat gAjghl"
+  class="sc-htpNat fxMOOa"
```
txwkx: 🤔 hmm

README.md
```diff
@@ -24,9 +24,7 @@ components that conform to Marshmallow's Design System - "_S'mores_".
 ## Install
 
 ```bash
-$ yarn add @mrshmllw/smores-react
```
txwkx: moving the project back to `npm` wouldn't prevent people from installing it using `yarn`

txwkx: shall we leave the **README.md** as it is?

src/CheckBox/__tests__/__snapshots__/CheckBox.js.snap
txwkx: need to tackle those fake snapshot changes asap

README.md
alexwllms: That makes sense, although we should punish them.

PR #86, opened 2019-10-17
Reset dropdown when the list changes

hansonhsc:
## What does this do?

Ensure the dropdown resets to initial state, when the dropdown changes. Previously, if the user selected item 3 in the list, then the list changes, item 3 will still be selected even though the value and label has completely changed, because it's a new list.

## What does it affect?

The dropdown control used in Agent Portal and Customer Account

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -1,4 +1,4 @@
-import React, {FC, useEffect} from 'react';
+import React, {FC, useEffect, useState} from 'react';
 import styled from 'styled-components';
 
 import {Text} from '../Text';
@@ -14,7 +14,7 @@ export type DropdownItem = {
 type Props = {
   /** ID, usually used for tests  */
   id: string;
-  /** className attribute to apply classses from props */
+  /** className attribute to apply classes from props */
   className?: string;
   /** label displayed above the dropdown  */
   label?: string;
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -52,13 +60,14 @@ export const Dropdown: FC<Props> = ({
         <Select
           id={id}
           disabled={disabled}
-          defaultValue={list.length === 1 ? String(list[0].value) : placeholder}
-          onChange={(e: React.FormEvent<HTMLSelectElement>) =>
-            onSelect(e.currentTarget.value)
-          }
+          onChange={(e: React.FormEvent<HTMLSelectElement>) => {
+            setValue(e.currentTarget.value);
+            onSelect(e.currentTarget.value);
```
txwkx: since those two lines are logically identical to the [45-46], can we create a function, with a single argument that calls ***seValue*** and ***onSelect*** ?

src/Dropdown/Dropdown.tsx
```diff
@@ -52,13 +60,14 @@ export const Dropdown: FC<Props> = ({
...
+          }}
           required
+          value={value}
         >
-          {list.length > 1 && (
+          {list.length !== 1 && (
```
txwkx: I'm trying to think of the UX in case the *list*'s length is `0`

hansonhsc: In fact, I think we might want to just remove that conditional rendering here, as there's no harm having a *hidden* placeholder when the length is 1 and the single *unhidden* item is auto-selected.

src/Dropdown/Dropdown.tsx
hansonhsc: Yep, will do

src/Dropdown/Dropdown.tsx
hansonhsc: The list can be empty in the Custom Car MTA flow, where dropdowns like engine capacity will be empty before the manufacturer and model has been selected.

PR #107, opened 2019-10-31
Add Textarea component

alexwllms:
## What does this do?

Add a `Textarea` component to the project.

## What does it affect?

- Add a `Textarea` component to the project
- Add stories for the default, error and disabled states
- Add tests for the component

## Any debts to pay?

- Need to bare in mind that when integrating into the Agent Portal we'll need to add in some missing positional styles

## Review Checklist

- [x] Added and/or modified tests for affected areas
- [x] Squashed any extraneous commit messages
- [x] Ready to review
- [ ] Ready to merge

### DIFF
--- a/src/Textarea/Textarea.stories.js
+++ b/src/Textarea/Textarea.stories.js
@@ -14,7 +14,6 @@ storiesOf('Textarea', module)
       onChange={() => {}}
       disabled={boolean('disabled', false)}
       placeholder="Here is some placeholder text."
-      className=""
     />
   ))
   .addWithJSX('disabled', () => (
@@ -26,7 +25,6 @@ storiesOf('Textarea', module)
       resize="none"
       disabled={boolean('disabled', true)}
       placeholder="Here is some placeholder text."
-      className=""
     />
   ))
   .addWithJSX('error', () => (
@@ -43,7 +41,6 @@ storiesOf('Textarea', module)
         'Something really quite terrible has gone wrong here!',
       )}
       placeholder={text('placeholder', 'Some lovely placeholder text!')}
-      className=""
     />
   ))
   .addWithJSX('functional', () => <Container />);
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
   ],
   "scripts": {
     "test": "jest --coverage",
-    "test:watch": "jest --watch",
+    "test:watch": "jest --watch --coverage",
```
txwkx: is this PR up to date with `master`?

src/Box/Box.tsx
```diff
@@ -7,7 +6,7 @@ import {space, SpaceProps} from '../utils/space';
 
 interface IBox extends SpaceProps, FlexProps, MeasureProps {}
 
-export const Box: FC = styled.div<IBox>`
+export const Box = styled.div<IBox>`
```
txwkx: same comment as above

src/Textarea/Textarea.stories.js
```diff
@@ -0,0 +1,50 @@
...
+  .addWithJSX('default', () => (
+    <Textarea
+      id="textarea_id"
+      label={text('label', 'Textarea label')}
+      value={text('value', '')}
+      resize="both"
+      onChange={() => {}}
+      disabled={boolean('disabled', false)}
+      placeholder="Here is some placeholder text."
+      className={text('Class name', '')}
```
txwkx: this **props** is required by `styled-components` only
I don't think we would have any use of explicitly setting it in the storybook

src/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,106 @@
...
+  /** used for label - input connection */
+  name?: string;
+  /** input value */
+  value: string;
+  /** error flag */
+  error?: boolean;
+  /** error text message */
+  errorMsg?: string;
+  /** Allow user to resize the textarea vertically and horizontally or not */
+  resize: 'none' | 'both';
```
txwkx: shall we set `resize` as an optional prop that defaults to ***none***?

the number of use cases of users adjusting the size in practice is extremely low (personal assumption)

package.json
alexwllms: Updated

PR #110, opened 2019-11-04
Proposal: update Prettier config and update options

alexwllms:
## What does this do?

This is a proposal to update Prettier config and update the config to add spaces around arguments inside brackets (sensible whitespace) and remove semi-colons (as the compiler will add these on build when minimised).

## What does it affect?

- Remove reference to config options that are the `default` for Prettier
- Set `"bracketSpacing": true` to add spacing around arguments
- Set `"semi": false,` to remove semi-colons

## Review Checklist

- [x] Added and/or modified tests for affected areas
- [x] Squashed any extraneous commit messages
- [x] Ready to review
- [x] Ready to merge

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,17 +1,7 @@
 {
-  "arrowParens": "avoid",
   "bracketSpacing": false,
-  "htmlWhitespaceSensitivity": "css",
-  "insertPragma": false,
-  "jsxBracketSameLine": false,
-  "jsxSingleQuote": false,
   "proseWrap": "always",
-  "quoteProps": "as-needed",
-  "requirePragma": false,
   "semi": true,
   "singleQuote": true,
-  "tabWidth": 2,
-  "trailingComma": "all",
-  "useTabs": false,
-  "printWidth": 80
+  "trailingComma": "all"
 }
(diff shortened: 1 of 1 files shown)


txwkx:
@alexwllms can you share some side-by-side screenshots of a file with different formatting?

really curious because the majority of **prettierrc** is gone 🙀

alexwllms:
@txwkx while the majority of the `.prettierrc` is gone the effect remains the same. All of the lines that I removed were the default settings for Prettier. So even though we don't set the options explicitly Prettier will still implement the default rule set.

For example: 

`"tabWidth": 2` is the default setting for Prettier. So we can remove the explicit option and Prettier will still set the tabWidth to 2 when formatting the file. We only need to explicitly set the options if we want the option to do something other than the default behaviour.

The only options we are setting that are not the default are:

```
  "bracketSpacing": true,	
  "proseWrap": "always",	
  "semi": false,	
  "singleQuote": true,	
  "trailingComma": "all"
```

The only two substantive changes from the previous config are to:

1. remove semicolons (screenshot incoming) and
2. add space around arguments in brackets (screenshot incoming).

alexwllms:
With semicolons and no-spacing in brackets:
before

Without semicolons and with spacing in brackets:
after

txwkx:
image

txwkx:
@alexwllms do you plan on formatting **all** files after the next release?

alexwllms:
@txwkx I suggest we reformat as we go along?

PR #411, opened 2020-04-23
Feature/Add Pound icon

MPoorter:
## Screenshot

image

## What does this do?

Add Pound icon

## What does it affect?

- Icons


src/Icon/Icon.tsx
```diff
@@ -430,6 +430,16 @@ export const Icon: FC<Props> = ({
         />
       </svg>
     )}
+
+    {// POUND
+    render === 'pound' && (
+      <svg width="16" height="16" viewBox="0 0 32 32" fill="none">
```
alexwllms, line 436: Set the width and height to 100% then it'll be scaleable when imported by other projects.

PR #461, opened 2020-05-22
🎹 implement ActionDropdown

txwkx:
## Screenshot

image

## What does this do?

- Introduce a new `<ActionDropdown />` component aka a "Rich" Dropdown with custom UI.

## Any debts to pay? (optional)

- This component isn't optimized for mobile devices

## Review Checklist

- [ ] Added and/or updated tests
- [ ] Ready to review
- [ ] Ready to merge

### DIFF
--- a/src/ActionDropdown/ActionDropdown.stories.js
+++ b/src/ActionDropdown/ActionDropdown.stories.js
@@ -1,0 +1,55 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+
+import { ActionDropdown } from './ActionDropdown'
+
+const defaultAction = {
+  label: 'Medium',
+  value: 'MEDIUM',
+  bgColor: 'yellow5',
+  textColor: 'blue7',
+}
+
+const actions = [
+  {
+    label: 'Default grey',
+    value: 'DEFAULT',
+    bgColor: 'grey2',
+    textColor: 'blue7',
+  },
+  {
+    label: 'High alert',
+    value: 'HIGH_ALERT',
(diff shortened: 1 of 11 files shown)


src/ActionDropdown/ActionDropdown.stories.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+
+import { ActionDropdown } from './ActionDropdown'
+
+const defaultAction = {
+  label: 'Medium',
```
alexwllms, line 7: This feels like a possible area for confusion, two strings that are uppercase and sentence case. Maybe turn the value into a `const`?

src/ActionDropdown/List.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { FC } from 'react'
+import styled from 'styled-components'
+
+import { theme } from '../theme'
+
+export type ActionListItem = {
+  label: string
+  value: string
```
alexwllms, line 8: Could this be a string literal to address my earlier comments?

txwkx: Answering to both comments here:

In order to avoid narrowing down the potential use cases for this component, I left the type definition quite generic.

It is up to the user to define what the _label-value_ pairs would look like.

e.g:
```js
{ label: "I am a label", value: "value"  }
or
{ label: "NoT RelAteD to Th3 Val", value: "foobar" }
```
@alexwllms

PR #656, opened 2020-09-24
Update theme.ts to include two new colours

ellajay:
## What does this do?

Updates theme.ts to include `cyan5` and `premfina orange`

## What does it affect?

- theme.ts


txwkx (approved):
🤙

PR #670, opened 2020-09-30
Migrate Storybook from version 5.3 to 6

alexwllms:
## What does this do?

Storybook v5.3 and more recently v6 introduces major breaking changes to how the config is set up and how individual stories are written. 

This PR migrates the config, and rewrites each story in the new format.

It also removes the `addons-knobs` which has been depreciated in favour of using the new controls package. 

## What does it affect?

- Replace all config files with new `main.js` file
- Remove redundant `codacy.yml` file
- Rewrite all stories in new module format
- Remove unnecessary packages

## Review Checklist

- [x] Added and/or updated tests
- [x] Ready to review
- [x] Ready to merge


package.json
```diff
@@ -58,17 +55,16 @@
     "jest": "^26.4.2",
     "jest-styled-components": "^7.0.3",
     "lint-staged": "^10.4.0",
-    "prettier": "^2.1.2",
-    "react-docgen-typescript-loader": "^3.7.2",
-    "storybook-addon-jsx": "^7.3.4"
+    "prettier": "^2.0.5",
```
ellajay, line 58: Is there an issue with the latest prettier release that made you change this to an earlier version?

alexwllms: Good catch. I was testing out a fix and forgot to revert it!

package.json
```diff
@@ -58,17 +55,16 @@
...
+    "react-docgen-typescript-loader": "^3.7.2"
   },
   "dependencies": {
     "date-fns": "^2.16.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "snyk": "^1.404.0",
```
lapostoj, line 65: Not fully relate but wouldn't snyk rather be a devDependencies? 🤔 not fully sure what it actually does as a JS dependency here...
Can also be ignore not to add risk to the PR as these are always tricky...

lapostoj (approved):
A bit from afar about the new stories setup but seems legit!
Good catch from Ella for prettier though yes

PR #759, opened 2020-11-04
Upgrade react version to 17.0.1

lapostoj:
I've basically broken the signup-flow by migrating to react 17 there to keep up with create-react-app.
And since this one is bringing in react 16.x it's conflicting. As I guess we'd want to update it everywhere anyway this can be a start.

To me that also bring the question if all the "dependencies" in there should be peerDependencies but that might need further change in setup so let's maybe keep it for later.

(I've also moved snyk as it's most likely just a dev dependency, right?)

Can you also remind me if anything specific would be needed to have it released or just need to merge it?


alexwllms (approved):
LGTM – let's see what we break!

PR #768, opened 2020-11-09
Modal component

habin-isa:
## Modal component added
- Takes title, children, handleClick (for toggle) and showModal
- Storybook example contains a button to show usage
- ⭐️ Adds drawer prop for drawer state on mobile view (Jason requested for this to be an optional prop)

image
image
image


src/Modal/Modal.stories.js
```diff
@@ -0,0 +1,34 @@
...
+    </Box>
+  )
+}
+
+export const Default = Template.bind({})
+
+Default.args = {
+  title: "Hello world i'm a beautiful modal",
+  showModal: false,
+}
```
alexwllms, line 34: Still not adding a newline on save? VS code settings? Prettier settings?

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export const Modal: FC<ModalProps> = ({
+  title = '',
+  body = '',
+  children = '',
+  showModal = false,
+  handleClick,
+}) => {
+
+  return (
```
alexwllms, line 30: If there's no body to the modal function can we remove the empty line and the return keyword and use the shorthand `}) => ...`

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <Wrapper showModal={showModal}>
+      <Overlay />
+      <Container>
+        <Box flex alignItems="center" justifyContent="space-between">
+          <Text tag="h2" typo="header-small" align="left">{title}</Text>
+          <IconContainer onClick={handleClick}>
```
alexwllms, line 36: Does Jason want us to use the new icon button with this modal (seen here https://www.figma.com/file/FqSaizGOyOzXFZNmZ4X0LGMn/%F0%9F%8D%AC-S-mores?node-id=2790%3A0)?

src/Modal/__tests__/Modal.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import {render} from '@testing-library/react';
```
alexwllms, line 2: Prettier doesn't seem to be adding spaces around the imports inside the `{ render }`

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+  body?: string,
+  children?: ReactNode,
+  showModal?: boolean,
+  handleClick: () => void
+}
+
+export const Modal: FC<ModalProps> = ({
+  title = '',
+  body = '',
+  children = '',
```
alexwllms, line 25: Should children be optional? I can see the modal having an optional title but I imagine it'll always required some child elements?

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Text } from '../Text'
+
+interface IModal {
+  // showModal state
+  showModal: boolean
+}
+
+type ModalProps = {
+  title?: string,
+  body?: string,
```
alexwllms, line 16: Where is the body used in the component? Am I being dumb 😆 I can't see it.

alexwllms (changes requested):

src/Modal/Modal.stories.js
habin-isa: Will have to check this!

src/Modal/Modal.tsx
habin-isa: Discussing now 😄

src/Modal/Modal.tsx
habin-isa: Oh no! Forgot to remove, me being dumb!

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,105 @@
...
+  showModal: boolean
+}
+interface IModalContainer {
+  // drawer state
+  drawer: boolean
+}
+
+type ModalProps = {
+  title?: string,
+  children?: ReactNode,
```
alexwllms, line 20: Should children be optional? It's optional here but has no default state when destructured in the component?

alexwllms (approved):

alexwllms:
@habin-isa one comment on the PR worth considering but then merge when you're ready!

PR #775, opened 2020-11-11
Fixes font error for ActionDropdown

habin-isa:
## Response to Joao's alert of dropdown font not rendering correctly
- Changes ActionDropdown + Text font to 'Circular'
- Updates snapshot tests

image

### DIFF
--- a/src/ActionDropdown/ActionDropdown.tsx
+++ b/src/ActionDropdown/ActionDropdown.tsx
@@ -6,7 +6,6 @@ import { Icon } from '../Icon'
 import { List, ActionListItem } from './List'
 
 import { theme } from '../theme'
-import { FontStyle } from '../fontStyle'
 
 type Props = {
   /** ID, usually used for tests  */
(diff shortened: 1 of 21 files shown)


joaocanaverde (approved):

src/ActionDropdown/ActionDropdown.tsx
```diff
@@ -56,6 +57,7 @@ export const ActionDropdown: FC<Props> = ({
       <OuterContainer open={open}>
         <List options={list} selectOption={onSelect} />
       </OuterContainer>
+      <FontStyle />
```
alexwllms, line 60: Should we load the font-face declarations into each component or add them globally?

alexwllms (changes requested):

src/ActionDropdown/ActionDropdown.tsx
habin-isa: Happy with either!

habin-isa: Added globally and removed font-family declarations on component level

PR #833, opened 2020-12-01
Adds new icons

habin-isa:
## Additional icons for sign-up

* doc
* breakdown
* car-mta


src/Icon/Icon.tsx
```diff
@@ -666,6 +666,203 @@ export const Icon: FC<Props> = ({
         </svg>
       )
     }
+
+    {
+      //CAR MTA
+      render === 'car-mta' && (
+        <svg
+          width="47"
```
alexwllms, line 674: Change the width and height to 100% (but leave the viewbox the same) and then each icon becomes scaleable!

src/Icon/Icon.tsx
```diff
@@ -666,6 +666,203 @@ export const Icon: FC<Props> = ({
...
+          height="47"
+          viewBox="0 0 47 47"
+          fill="none"
+          xmlns="http://www.w3.org/2000/svg"
```
alexwllms, line 678: This can be removed from all SVGs as it's redundant.

src/Icon/Icon.tsx
```diff
@@ -666,6 +666,203 @@ export const Icon: FC<Props> = ({
...
+        >
+          <path
+            d="M19.0938 38.1875C19.0938 40.621 17.121 42.5938 14.6875 42.5938C12.254 42.5938 10.2812 40.621 10.2812 38.1875C10.2812 35.754 12.254 33.7812 14.6875 33.7812C17.121 33.7812 19.0938 35.754 19.0938 38.1875Z"
+            fill="#112035"
```
alexwllms, line 682: Is there an expectation that this icon will change colour? If so use `fill={theme.colors[color]}` for the fill of the colour you need to be made dynamic.

alexwllms (changes requested):
A few comments that apply to all icons

src/Icon/Icon.tsx
habin-isa: Confirmed with Georgia that this won't be an expectation

alexwllms (approved):
👍

PR #870, opened 2020-12-16
added width prop to modal for refactoring

liampuk:
## Screenshot

image

## What does this do?

Modal width can now be set with an optional prop. This enables refactoring the old Modal in customer-portal-www.

## What does it affect?

- Modal


habin-isa (approved):
:shipit:

PR #871, opened 2020-12-16
added icon prop to modal

liampuk:
## Screenshot

image

## What does this do?

Added an optional icon prop to modal

## What does it affect?

- Modal


habin-isa (approved):
:shipit:

PR #892, opened 2021-01-05
use flex to properly center modal on scrolling pages

liampuk:
## What does this do?

Using flex for centring the modal horizontally and vertically, to ensure it is properly centred on pages longer than the viewport.   

## What does it affect?

- Modal


habin-isa (approved):
:shipit:

PR #893, opened 2021-01-06
🐛 removed hardcoded flex from modal

liampuk:
## What does this do?

Currently Modals will show by default, as the variable to show/hide is ignored by setting display twice.


habin-isa (approved):
:shipit:

PR #898, opened 2021-01-11
Added Emoji Component

fleonard:
## Screenshot

Screenshot 2021-01-11 at 13 27 55

## What does this do?

Added new component to properly render an emoji into a span with option to change size and add an a11y friendly label

### DIFF
--- a/src/Emoji/Emoji.tsx
+++ b/src/Emoji/Emoji.tsx
@@ -1,0 +1,60 @@
+import React, { FC } from 'react'
+import styled, { css } from 'styled-components'
+
+type EmojiProps = {
+  /** className attribute to apply classes from props */
+  className?: string
+  /** Emoji symbol */
+  symbol: string
+  /** set size of the Emoji  */
+  size?: number
+  /** aria-label to identify Emoji */
+  label?: string
+  /** margins TOP-RIGHT-BOTTOM-LEFT */
+  mt?: string
+  mr?: string
+  mb?: string
+  ml?: string
+}
+
+export const Emoji: FC<EmojiProps> = ({
+  className = '',
+  symbol,
+  size = 12,
+  label = 'emoji',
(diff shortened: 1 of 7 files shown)


src/Emoji/Emoji.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type EmojiProps = {
+  /** className attribute to apply classes from props */
+  className?: string
+  /** Emoji symbol */
+  symbol: string
+  /** set size of the Emoji  */
+  size?: number
+  /** aria-label to identify Emoji */
+  label?: string
```
alexwllms, line 12: Do you think this should be enforced rather than optional to avoid a bunch of "emoji" strings?

src/Emoji/README.md
```diff
@@ -0,0 +1,23 @@
+# < Emoji />
+
+## Props:
+
+```ts
+symbol:   string;
```
alexwllms, line 6: Should we clarify that you have to add (or copy and paste) the emoji in directly as a string rather than it's text description?

src/Icon/README.md
```diff
@@ -65,4 +65,4 @@ const App = () => (
 * clock
 * up-down
 * upload-email
-* warning
\ No newline at end of file
+* warning
```
alexwllms, line 68: Just checking that you meant to do the changes in this file as they look unrelated? If so ignore me.

alexwllms (approved):
A few comments but well done! You are doing god's work @fleonard

src/Icon/README.md
fleonard: yeah I've just updated the comment since it was a copy paste from the Box component so I've updated it

src/Emoji/README.md
fleonard: Good point

PR #900, opened 2021-01-11
Add flex grow to children container button

MPoorter:
## Screenshot

image

image

image

## What does this do?

Adding flex grow 1 to children of button so that the the icon is kept to the left and text is centred, at Jason's request

## What does it affect?

- Buttons buttons butttons


habin-isa (approved):
:shipit:

PR #901, opened 2021-01-11
Date picker shift days ignore weekends

liampuk:
## Screenshot

image

## What does this do?

- Added optional Props to DatePicker to disable picking weekends, and to shift the days along correctly (like a calendar). Shifting the days is necessary as otherwise weekends are in the middle sometimes (looks very odd)
- caret icons aren't displaying properly without adding !important to the padding on the styling here.

**I think shifting the days along correctly would be nice to have by default, but didn't want to break any existing layouts**

## What does it affect?

- Date Picker


habin-isa (approved):
:shipit:

PR #908, opened 2021-01-12
Disable button when loading and hide icon when loading

MPoorter:
## Screenshot

https://user-images.githubusercontent.com/25356097/104315407-cf236200-54d2-11eb-8081-68291f462379.mov

## What does this do?

Disables the button when loading to prevent double click
Hides the icon as well when loading

## What does it affect?

- More buttons


habin-isa (approved):
:shipit:

PR #913, opened 2021-01-14
shift sunday correctly when first day of month

liampuk:
## Screenshot

bug: 
image

fix:
image

## What does this do?

Fixes a bug where calendar dates are not shifted correctly when the first day of the month is a Sunday

## What does it affect?

- Date picker


habin-isa (approved):
:shipit:

PR #970, opened 2021-02-12
Change Tag font to Circular

MPoorter:


src/Tag/Tag.tsx
```diff
@@ -40,6 +40,7 @@ const Wrapper = styled.div<IWrapper>`
 `
 
 const TagText = styled(Text)`
-  font-family: 'Gordita', sans-serif;
+  font-family: 'Circular', sans-serif;
```
fleonard, line 43: I think you can just get rid of this since `Circular` should be the default font family anyway

fleonard (changes requested):

fleonard (approved):

PR #1053, opened 2021-03-17
Et 3073 row component

ellajay:
## Screenshot / video

image

## What does this do?

Adds the Row component and related tests.


src/Row/Row.tsx
```diff
@@ -0,0 +1,56 @@
...
+interface IContainer {
+  type?: 'first' | 'last' | 'middle' | 'curved'
+}
+
+const Container = styled.div<IContainer>(
+  ({ type }) => css`
+    border-radius: ${(type === 'first' && `8px 8px 0 0`) ||
+    (type === 'curved' && `8px`) ||
+    (type === 'last' && `0 0 8px 8px`) ||
+    (type === 'middle' && 0)};
```
fleonard, line 49: Do we really need the `middle` type?
We could probably get rid of it and maybe have `border-radius:0` as default fallback in case no type is passed?

src/Row/Row.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const Row: FC<Props> = ({
+  icon,
+  heading,
+  subHeading,
+  type,
+  caret = true,
+}) => {
+  return (
+    <Container type={type}>
+      <Box>{icon && <Icon render={icon} size={24} />}</Box>
```
fleonard, line 26: I'm kinda wondering if instead of having a fixed grid-template and render empty divs to make them fill the 24px space of the template we should probably have dynamic grid-templates based on icon and caret and just do something like:
```
{icon && <Icon render={icon} size={24} />}
...
{caret && <Icon render="caret" rotate={-90} color="grey8" />}
```

What do you think?

fleonard (changes requested):
Left you a couple of thoughts

src/Row/Row.tsx
ellajay: True!

src/Row/Row.stories.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react'
+import { Row } from '../Row'
+
+export default {
+  title: 'Row',
+  conmponent: Row,
```
fleonard, line 6:
```suggestion
  component: Row,
```

src/Row/Row.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+import { theme } from '../theme'
+import { Icon } from '../Icon'
+import { Box } from '../Box'
+import { Text } from '../Text'
+
+export type Props = {
+  iconLeft?: string
+  iconRight?: string
+  iconRightColor?: string
```
fleonard, line 12: Should we add color also for the left one or do you think that will always be default?

src/Row/Row.tsx
```diff
@@ -0,0 +1,95 @@
...
+    (type === 'curved' && `8px`) ||
+    (type === 'last' && `0 0 8px 8px`) ||
+    0};
+    border: 1px solid ${theme.colors.grey4};
+    display: grid;
+    grid-template-columns: ${iconLeft ? `5% auto 5%` : `auto 5%`};
+    padding: 24px;
+    align-items: center;
+    border-top: ${borderTop === false && `none`};
+    border-bottom: ${borderBottom === false && `none`};
```
fleonard, line 83:
```suggestion
    border-top: ${!borderTop && `none`};
    border-bottom: ${!borderBottom && `none`};
```

src/Row/Row.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    .iconRight {
+      justify-self: end;
+    }
+
+    @media (max-width: 768px) {
+      grid-template-columns: ${iconLeft ? `10% auto 10%` : `auto 10%`};
+      padding: 16px;
+      grid-gap: 12px;
+    }
```
fleonard, line 93: Can we convert this to be `min-width` in order to style always mobile first?

fleonard (changes requested):

src/Row/Row.tsx
ellajay: May as well!

fleonard (approved):

PR #1083, opened 2021-03-30
Row grid gap

ellajay:
## Screenshot / video

image

## What does this do?

Adds an optional `boldHeading` prop to the Row component and updates the grid-gap.
Also adds the new `claim`, `no-excess` and `tow-truck` icons to the Icon component.


src/Row/Row.tsx
```diff
@@ -68,31 +71,36 @@ interface IContainer {
...
+      font-weight: ${boldHeading && 500};
+    }
+
     .iconRight {
       justify-self: end;
     }
 
     @media (min-width: 768px) {
-      grid-template-columns: ${iconLeft ? `5% auto 5%` : `auto 5%`};
+      grid-template-columns: ${iconLeft ? `24px auto 24px` : `auto 24px`};
```
fleonard, line 101: this is the same as mobile now so you can remove it

```suggestion

```

fleonard (changes requested):
Nice one! Just a very minor css update and it's good to go!

src/Row/Row.tsx
ellajay: Good point! Thank you :)

fleonard (approved):

PR #1089, opened 2021-04-01
Added release workflow

fleonard:
## What does this do?

After this PR is merged, every tag push will trigger the new release workflow that will create a release and populate it from `CHANGELOG.md`

Once this is released we need to remember to populate `CHANGELOG.md` with what we are releasing either during release process or as part of the branch with the new feature or fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+## [0.60.0] - 2020-04-01
+### Added
+- Added new release tag workflow
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+## [0.60.0] - TBC
```
fleonard, line 1: I will update the date once I'm gonna release

habin-isa (approved):
Wow okkkk!! :shipit:

PR #1115, opened 2021-04-09
Et 3174 icon size

ellajay:
## Screenshot / video

image

## What does this do?

Updates the `Info`, `Cross` and `Warning` Icons to use more recent svgs. 
Minor adjustments to the `Row` component, including padding, grid-gap and conditional sizing of `iconLeft` depending on the screen width.


CHANGELOG.md
```diff
@@ -1,13 +1,19 @@
-## [0.60.0] - 2020-04-07
+## [0.60.1] - 2021-04-09
+### Updated
```
fleonard, line 2: If we follow `keepachangelog` format this can either be Added/Removed/Changed/Fixed so in this case I suppose can all go under Changed

fleonard (changes requested):

fleonard (approved):

PR #1239, opened 2021-06-14
Removed Figma workflow

fleonard:
I tested it a couple of times and doesn't seem to work at all.
Since it's not maintained anymore there's no point to insist with it, I'll find another way to do it.


ellajay (approved):

habin-isa (approved):
What a shame! :shipit:

PR #1272, opened 2021-07-02
Add modal classname

ellajay:
## What does this do?

Adds a `containerClass` to the Modal container so that styles are extendable via `className`


habin-isa (approved):
:shipit:

fleonard (approved):

PR #1327, opened 2021-07-29
Generate icons from figma

fleonard:
## Screenshot / video

Just a snipped of some of the new icons

Screenshot 2021-07-29 at 12 02 29

## What does this do?

I've created a node script to generate a map of all the icons that we can export from Figma and use that map to generate the Icon component. Functionality of the component won't change, we still use the same props as before but all the icons has been resized in Figma so we need to think of a way to release this because it might break some things.

I'll run through this PR and the script to export the icons from Figma in the next FE guild


src/assets/icons/iconMap.js
```diff
@@ -0,0 +1,349 @@
+import { ReactComponent as Aa } from './aa.svg'
+import { ReactComponent as AccountBreakdown } from './account-breakdown.svg'
+import { ReactComponent as AccountMlp } from './account-mlp.svg'
+import { ReactComponent as ActivityClipboard } from './activity-clipboard.svg'
+import { ReactComponent as AddOns } from './add-ons.svg'
+import { ReactComponent as Addons } from './addons.svg'
```
ellajay, line 6: What's the difference between these two different svgs?

src/assets/icons/iconMap.js
```diff
@@ -0,0 +1,349 @@
...
+import { ReactComponent as Fire } from './fire.svg'
+import { ReactComponent as Flag } from './flag.svg'
+import { ReactComponent as Folder } from './folder.svg'
+import { ReactComponent as FraudAlert } from './fraud-alert.svg'
+import { ReactComponent as Fuel } from './fuel.svg'
+import { ReactComponent as Globe } from './globe.svg'
+import { ReactComponent as GooglePlayBadge } from './google-play-badge.svg'
+import { ReactComponent as Gpay } from './gpay.svg'
+import { ReactComponent as HelpAboutus } from './help-aboutus.svg'
+import { ReactComponent as HelpBuypolicy } from './help-buypolicy.svg'
```
ellajay, line 74: Very minor but should this be `HelpBuyPolicy` in line with the format of `HelpManagePolicy` below?

src/assets/icons/iconMap.js
```diff
@@ -0,0 +1,349 @@
...
+import { ReactComponent as HelpCenter } from './help-center.svg'
+import { ReactComponent as HelpIdcheck } from './help-idcheck.svg'
```
ellajay, line 76: Same here - should this be `HelpIdCheck`?

ellajay (approved):

src/assets/icons/iconMap.js
fleonard: Apparently one has some pink and the other doesn't 😁

src/assets/icons/iconMap.js
fleonard: This is all generated by the script based on the component names we have on Figma so the PascalCase convert based on where the hyphen is. It doesn't matter that much considering that the name we need to use is the one in Figma anyway

src/assets/icons/iconMap.js
ellajay: Good to know!

trcywu (approved):

PR #1328, opened 2021-07-29
Message text

habin-isa:
## What does this do?
- Adds `sizeSmall` prop as on sign up we need the icon size and font size to be smaller
- Updates changelog

### DIFF
--- a/src/Message/Message.tsx
+++ b/src/Message/Message.tsx
@@ -10,6 +10,7 @@ export type MessageProps = {
   type?: 'info' | 'warning' | 'warning-bubble'
   alignIcon?: 'center' | 'flex-start' | 'flex-end'
   backgroundColor?: string
+  sizeSmall?: boolean
 }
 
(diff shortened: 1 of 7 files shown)


src/Message/Message.tsx
```diff
@@ -10,17 +10,24 @@ export type MessageProps = {
...
 export const Message: FC<MessageProps> = ({
   children,
   type = 'info',
   alignIcon = 'center',
   backgroundColor,
+  sizeSmall,
 }) => (
-  <Wrapper type={type} backgroundColor={backgroundColor}>
+  <Wrapper type={type} backgroundColor={backgroundColor} sizeSmall={sizeSmall}>
+    {sizeSmall}
```
fleonard, line 24: this looks odd placed here

habin-isa: Definite error not sure how that got there lol, will remove

fleonard (approved):

PR #1424, opened 2021-09-16
adds extra props and styles to search input, updates snapshots

JamieMarsay:
Adds props and styling that allows the SearchInput component to be used in the new sign up flow on


src/Icon/Icon.tsx
```diff
@@ -3650,6 +3650,20 @@ export const Icon: FC<Props> = ({
...
+        viewBox="0 0 12 8"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+      >
+        <path
+          d="M11.7444 0.18579C11.4316 -0.0864597 10.9591 -0.0549597 10.6861 0.25704L6.00012 5.61129L1.31412 0.25629C1.04112 -0.0557097 0.567867 -0.0872097 0.255867 0.18579C-0.0561332 0.45879 -0.0868832 0.932791 0.186117 1.24404L5.43612 7.24404C5.57862 7.40754 5.78337 7.50054 6.00012 7.50054C6.21687 7.50054 6.42162 7.40754 6.56412 7.24404L11.8141 1.24404C12.0871 0.932791 12.0556 0.45879 11.7444 0.18579Z"
+          fill="#D2D2D2"
+        />
+      </svg>
+    )}
```
fleonard, line 3666: you don't really need a new icon for this, we already have `caret`, you can just rotate it in the Icon component

src/SearchInput/SearchInput.tsx
```diff
@@ -11,19 +12,29 @@ export type SearchInputItem = {
 
 interface IContainer {
   show: boolean
+  hasBorder?: boolean
+}
```
fleonard, line 16: you can just extend `IInnerContainer` interface instead of adding the props to all the others so it looks more clean

src/SearchInput/SearchInput.tsx
```diff
@@ -41,6 +52,10 @@ type SearchInputProps = {
   onFound: (element: string) => void
   /** Displays search results in a relative position to other elements */
   resultsRelativePosition?: boolean
+  /** Displays caret */
+  showCaret?: boolean
+  /** Displays border */
+  hasBorder?: boolean
```
fleonard, line 58: can you call this `outlined` so it's in line with the same prop we already have in other components?

src/SearchInput/SearchInput.tsx
```diff
@@ -51,6 +66,8 @@ export const SearchInput: FC<SearchInputProps> = ({
   searchList,
   onFound,
   resultsRelativePosition = false,
+  showCaret,
+  hasBorder,
```
fleonard, line 70: you should default them to false so they don't break any existing functionality

src/SearchInput/SearchInput.tsx
```diff
@@ -130,11 +158,22 @@ const Container = styled.div<IContainer>`
...
+  ${(p) =>
+    p.hasBorder
+      ? `
+    border: 1px solid ${theme.colors.grey3};
+    border-radius: 8px;
+    padding: 16px;
+  `
+      : ``}
+`
+
```
fleonard, line 171: shouldn't this has position relative otherwise the icon will be positioned relative to the parent

src/SearchInput/SearchInput.tsx
```diff
@@ -93,33 +110,44 @@ export const SearchInput: FC<SearchInputProps> = ({
...
+          onChange={updateInputState}
+          hasBorder={hasBorder}
+        />
+
+        {showCaret && (
+          <IconContainer>
+            <Icon render="caret-down" />
+          </IconContainer>
+        )}
+
```
fleonard, line 132: shouldn't you close Icon Container here?

fleonard (commented):
Left some comments, wondering if it's also worth doing it for the dropdown

Just a reminder you also need to update the changelog file before releasing

For next PR, can you also attach a screenshot of the outcome so it will be easier for me to review the changes and have a better context

src/SearchInput/SearchInput.tsx
JamieMarsay: Not sure what you mean... it has a closing tag

src/SearchInput/SearchInput.tsx
habin-isa: This can be shortened by using `&&` instead of a `?` => `p.hasBorder && border: 1px solid ${theme.colors.grey3};` etc

src/SearchInput/SearchInput.tsx
fleonard: sorry I ment the InnerContainer. I don't think it should wrap the results. As far as I saw it the inner container should just be a wrapper for the input and the icon, right?

habin-isa:
As it's a `tsx` file it'd be worth switching to importing `css` from styled-components as it's cleaner. Then you can follow this type of format instead of using `p.hasBorder` everywhere.
```
const InnerContainer = styled.div<IInnerContainer>(
  ({ hasBorder }) => css`
    ${hasBorder && `border: 1px solid ${theme.colors.grey3};`}
  `
)
```

src/SearchInput/SearchInput.tsx
habin-isa: Please add an explicit size prop declaration for the Icon 😎

src/SearchInput/SearchInput.tsx
JamieMarsay: so i've made the props optional in TS and then if they're not passed through they'll be evaluated as falsey anyway, so not sure they need the default prop putting on, but happy to if it's must :)

habin-isa (changes requested):
image
The caret is not aligned and the colour text and spacing does not match the designs as below:
image

src/SearchInput/SearchInput.tsx
```diff
@@ -11,19 +12,29 @@ export type SearchInputItem = {
 
 interface IContainer {
```
habin-isa, line 13: This one doesn't look like it's used for anything so can remove here and in the styled component `Container`

JamieMarsay:
@habin-isa Where are you seeing the issue with the caret alignment? 

Have fixed colour and sizing 👍 

Screenshot 2021-09-16 at 16 33 48

habin-isa (commented):
Looks good - what was the reasoning behind switching to `outlined`? Curious as `hasBorder` naming is used in other smores components

src/SearchInput/SearchInput.stories.js
```diff
@@ -9,4 +9,18 @@ export default {
...
+    id="days"
+    label="Days"
+    placeholder="Search days"
+    searchList={list}
+    onFound={() => {}}
+    showCaret
+    outlined
+  />
+)
+
```
fleonard, line 23: instead of duplicating this one you should have added the props in `Container.tsx` so if we even need to add others we can just expand on that instead of creating duplicates of the same component

src/SearchInput/SearchInput.tsx
```diff
@@ -51,6 +56,8 @@ export const SearchInput: FC<SearchInputProps> = ({
   searchList,
   onFound,
   resultsRelativePosition = false,
+  showCaret,
+  outlined,
```
fleonard, line 60: those need to be defaulted to false

src/SearchInput/SearchInput.tsx
```diff
@@ -124,51 +142,74 @@ export const SearchInput: FC<SearchInputProps> = ({
...
+
 const ResultsContainer = styled.div<IResultsContainer>`
   box-sizing: border-box;
   overflow-y: hidden;
-  ${(p) => p.absolutePosition && 'position: absolute;'}
+  ${({ absolutePosition }) => absolutePosition && 'position: absolute;'}
   width: 100%;
-  visibility: ${(p) => (p.show ? 'visible' : 'hidden')};
+  visibility: ${({ show }) => (show ? 'visible' : 'hidden')};
+  ${({ outlined }) => outlined && 'left: 0px; top: 90%;'}
```
fleonard, line 194: any reason why this is needed? It's hard to understand styling changes without a video of the functionality of it

src/SearchInput/SearchInput.tsx
```diff
@@ -124,51 +142,74 @@ export const SearchInput: FC<SearchInputProps> = ({
...
+      outlined ? theme.colors.grey8 : theme.colors.grey4};
   }
 `
 
+const IconContainer = styled.div`
+  position: absolute;
+  top: 50%;
+  right: 20px;
+  transform: translateY(-50%);
+`
```
fleonard, line 186: this didn't changed from the previous review  where @habin-isa pointed out that the caret wasn't aligned. Is it still the case?

fleonard (changes requested):
You are still missing the changes to the changelog file needed to release and the same updates for the dropdown component since looks like it's also needed for the NCD block

fleonard:
> Looks good - what was the reasoning behind switching to `outlined`? Curious as `hasBorder` naming is used in other smores components

@habin-isa we used outlined for the text input style with full border. hasBorder is not really meaningful since you don't know which kind of border it has while outlined is more clear to have it all around

JamieMarsay:
@fleonard 

In order of your comments (GitHub won't let me reply to each comment individually for some reason) - 

**Release/changelog comment:** 

So do the change log updates need to happen on my branch then? 

The readme suggests that I have to first checkout master and make the changes on there instead of on my branch...

Does the readme need updating to reflect this so that the process is clearer? 

**DropDown component changes:**

Happy to make them, had missed you mentioning this in the wider context of the comments received

**Container duplication:**

I duplicated this in the stories file because, for some reason, I couldn't pass my props following the existing .args pattern, so for the sake of time I went down this path to get things moving along as this work has taken a whole week so far. 

**Defaulting props to false:**

I had mentioned this in your previous comment, React will default props to null when they're not passed in which will then be evaluated as falsey, as they are new props added to the component there's no need to default them to false as that is already handled for us without the extra overhead. 

**Top/left styles neccesary:** 

I used the left and top properties to ensure that the results container is position correctly and hides the bottom of the input's border, otherwise there's misalignment and strange visual bugs with the results container, probably not an ideal fix, but again for the sake of time, I put this in and ran with it. 

**Icon Container styles:**

So I couldn't replicate the issue that Habs was having with the caret and also posted a screenshot showing it working, so haven't made any changes here because, as far as I can see, it all looks to be working fine.

fleonard:
@JamieMarsay, don't worry about the storybook container I pushed a fix for it and it's now working fine, you just needed to pass down the props and render them in the searchInput to make it works.

Changelog changes can be either made on your branch if you know that you need to release quickly or in a different branch, as long as the updated file is in master before you bump version and push tags it's fine

I have some concern about the caret icon for the `SearchInput` component. To me that layout looks more the one of a dropdown rather than a search input cause it prompt the user to click on it and expect a list of items while search input doesn't work this way. Can you please double check with design on what it's currently using and what we wanna do about it?
In my opinion the `SearchInput` component should just have a full outline without the icon, if the design prompt the icon then we should just change component and use the dropdown instead, even tho it could be problematic because the list of country might be a bit too long for it.

src/Dropdown/Dropdown.tsx
```diff
@@ -39,6 +45,7 @@ export const Dropdown: FC<Props> = ({
   disabled = false,
   list,
   onSelect,
+  outlined,
```
fleonard, line 48: can you default this to false please?

src/SearchInput/SearchInput.tsx
```diff
@@ -51,6 +53,7 @@ export const SearchInput: FC<SearchInputProps> = ({
   searchList,
   onFound,
   resultsRelativePosition = false,
+  outlined,
```
fleonard, line 56: same here

fleonard (commented):
just a couple of minor things and it's good to go

habin-isa (approved):
Good to go after Leo's comments are addressed :shipit:

fleonard (approved):
👍🏻

PR #1454, opened 2021-09-28
Add maxLength prop to Textarea

ellajay:
## What does this do?

Adds `maxLength` prop to Textarea component

### DIFF
--- a/src/Textarea/Textarea.tsx
+++ b/src/Textarea/Textarea.tsx
@@ -49,9 +49,11 @@ export const Textarea: FC<TextareaProps> = ({
 }) => (
   <Box flex direction="column" className={className}>
     {label && (
-      <Text tag="label" color="grey8" typo="label">
-        {label}
-      </Text>
+      <Box mb="4px">
+        <Text tag="label" color="grey8" typo="label">
+          {label}
+        </Text>
+      </Box>
     )}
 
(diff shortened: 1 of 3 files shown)


fleonard:
@ellajay Any chance you can also update the styling to mimic how the outlined textInput works?
After having a chat with Jason we decided to have the outline grey6 when the input is not empty, on hover and on focus and have a 2px border and I think we might also need to add 4px margin between the label and the textArea.

src/Textarea/Textarea.tsx
```diff
@@ -89,14 +96,23 @@ const Field = styled.textarea<ITextarea>`
...
 
   &:hover {
     border-color: ${({ error }) => theme.colors[`${error ? 'red7' : 'grey6'}`]};
   }
 
-  &:focus {
-    border-color: ${({ error }) => theme.colors[`${error ? 'red7' : 'blue5'}`]};
+  &:focus,
+  &:focus-visible {
+    border-color: ${({ error }) => theme.colors[`${error ? 'red7' : 'grey6'}`]};
```
fleonard, line 107: can you add those two rules with the hover one? they are basically the same

fleonard (approved):
Minor style update and it's good to go!
Remember to also add a new entry in the changelog file for release notes

PR #1462, opened 2021-09-29
added on change input handler

zakmarshmallow:
added two new handlers to TextInput onInputChange and onBlur

### DIFF
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -7,7 +7,7 @@ import { Icon } from '../Icon'
 
 import { theme } from '../theme'
 
-type Props = {
+type DefaultProps = {
   /** ID, usually used for tests  */
   id: string
   /** className attribute to apply classses from props */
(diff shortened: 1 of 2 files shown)


src/TextInput/TextInput.tsx
```diff
@@ -28,14 +28,26 @@ type Props = {
   error?: boolean
   /** error text message */
   errorMsg?: string
-  /** onChange listener */
-  onChange: (e: string) => void
+  /** onBlur listener */
+  onBlur?: (e: FormEvent<HTMLInputElement>) => void
   /** onChange listener */
```
fleonard, line 33: can you change this comment to describe the `trailingIcon` instead?

src/TextInput/TextInput.tsx
```diff
@@ -28,14 +28,26 @@ type Props = {
...
 
+type TruncateProps =
+  | {
+      onChange: (e: string) => void
+      onInputChange?: (e: FormEvent<HTMLInputElement>) => void
+    }
+  | {
+      onChange?: (e: string) => void
+      onInputChange: (e: FormEvent<HTMLInputElement>) => void
+    }
```
fleonard, line 47: Wondering if it's worth adding a comment to explain what's the difference and how both are used

src/TextInput/TextInput.tsx
```diff
@@ -70,9 +84,14 @@ export const TextInput: FC<Props> = ({
...
-        onChange={(e: FormEvent<HTMLInputElement>) =>
-          onChange(e.currentTarget.value)
-        }
+        onChange={(e: FormEvent<HTMLInputElement>) => {
+          if (onChange) onChange(e.currentTarget.value)
+
+          if (onInputChange) onInputChange(e)
+        }}
+        onBlur={(e) => {
+          if (onBlur) onBlur(e)
```
fleonard, line 93: I think `&&` instead of if might be a bit cleaner

src/TextInput/TextInput.tsx
```diff
@@ -70,9 +84,14 @@ export const TextInput: FC<Props> = ({
         value={value}
         error={error}
         autoComplete="off"
-        onChange={(e: FormEvent<HTMLInputElement>) =>
-          onChange(e.currentTarget.value)
-        }
+        onChange={(e: FormEvent<HTMLInputElement>) => {
+          if (onChange) onChange(e.currentTarget.value)
+
+          if (onInputChange) onInputChange(e)
```
fleonard, line 90: Shouldn't this be an if-else considering that we cannot have both of them?

fleonard (commented):
Just some very minor comments

fleonard (approved):
👍🏻

PR #1558, opened 2021-11-08
Styling fixes + extensions to Confirmation radio buttons

josianed:
## What does this do?

- fix placeholder text colour in text input and area components
- Make entire Confirmation radio buttons text selectable (rather than just the radio button itself)
- Make Confirmation radio buttons unselectable

### DIFF
--- a/src/ConfirmationRadioButtons/Confirmation.tsx
+++ b/src/ConfirmationRadioButtons/Confirmation.tsx
@@ -29,45 +29,59 @@ export const Confirmation: FC<ConfirmationProps> = ({
   sublabel,
   yesLabel = 'Yes',
   noLabel = 'No',
-}: ConfirmationProps) => {
-  return (
-    <ConfirmationWrapper>
-      <TextWrapper>
-        <SectionHeadingText tag="h3">{label}</SectionHeadingText>
-        {sublabel && (
-          <Text tag="p" typo="base-small" color={theme.colors.grey8}>
-            {sublabel}
-          </Text>
-        )}
-      </TextWrapper>
-      <RadioButtonGroupWrapper>
-        <RadioButtonGroup>
-          <RadioButtonWrapper checked={checked === true} error={error}>
-            <RadioButton
-              id={id}
-              label={yesLabel}
-              checked={checked === true}
-              onChange={() => onChange(true)}
-              value={id}
-              onBlur={onBlur}
-            />
-          </RadioButtonWrapper>
-          <RadioButtonWrapper checked={checked === false} error={error}>
-            <RadioButton
-              id={`${id}-1`}
-              label={noLabel}
-              checked={checked === false}
-              onChange={() => onChange(false)}
-              value={`${id}-1`}
-              onBlur={onBlur}
-            />
-          </RadioButtonWrapper>
-        </RadioButtonGroup>
-        {error && <ErrorBox>{errorMsg}</ErrorBox>}
-      </RadioButtonGroupWrapper>
-    </ConfirmationWrapper>
-  )
-}
+}: ConfirmationProps) => (
+  <ConfirmationWrapper>
+    <TextWrapper>
+      <SectionHeadingText tag="h3">{label}</SectionHeadingText>
+      {sublabel && (
+        <Text tag="p" typo="base-small" color={theme.colors.grey8}>
+          {sublabel}
+        </Text>
+      )}
+    </TextWrapper>
+    <RadioButtonGroupWrapper>
+      <RadioButtonGroup>
+        <RadioButtonWrapper
+          checked={checked === true}
+          error={error}
+          onClick={() =>
+            checked === true ? onChange(undefined) : onChange(true)
+          }
+        >
+          <RadioButton
+            id={id}
+            label={yesLabel}
+            checked={checked === true}
+            onChange={() =>
+              checked === true ? onChange(undefined) : onChange(true)
+            }
+            value={id}
+            onBlur={onBlur}
+          />
+        </RadioButtonWrapper>
+        <RadioButtonWrapper
+          checked={checked === false}
+          error={error}
+          onClick={() =>
+            checked === false ? onChange(undefined) : onChange(false)
+          }
+        >
+          <RadioButton
+            id={`${id}-1`}
+            label={noLabel}
+            checked={checked === false}
+            onChange={() =>
+              checked === false ? onChange(undefined) : onChange(false)
+            }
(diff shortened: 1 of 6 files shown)


src/ConfirmationRadioButtons/Confirmation.tsx
```diff
@@ -29,45 +29,59 @@ export const Confirmation: FC<ConfirmationProps> = ({
...
+      )}
+    </TextWrapper>
+    <RadioButtonGroupWrapper>
+      <RadioButtonGroup>
+        <RadioButtonWrapper
+          checked={checked === true}
+          error={error}
+          onClick={() =>
+            checked === true ? onChange(undefined) : onChange(true)
+          }
```
Jmuskett, line 49: So looking at this, I think my feedback is - I think we can simplify the component significantly, by removing the internal `checked` and `onClick` logic.   

As an example, in agent portal, in PolicyHolderInfo.tsx - starting on line 94, we have this 

```
    <ConfirmationWrapper>
          <Confirmation
            id="first-name"
            checked={fieldValues?.firstName}
            onChange={(val) => setFieldValues({ ...fieldValues, firstName: val })}
          />
        </ConfirmationWrapper>
``` 

if you change the `checked` prop to something like `hasFieldValue={fieldValues?.firstName}` 

then, in the smores component you could remove the onChange completely and change the checked on the yes/no buttons to 

```
checked={hasFieldValue}
```

```
checked={!hasFieldValue}
```

respectively, so the checkboxes would be controlled completely from outside of the component, and the actual checkbox component would just be concerned with presentation. You can then use styled components in the way you have, to apply the border etc.  

let me know if i'm missing any context - or thoughts on this?

jsudufour: I agree that the component should ideally be controlled entirely from outside the component, and think we should make some changes here if we can accommodate that. The caveat in this case is that when using this component with react-hook-form's `Controller` component, we need to pass in the provided `value` and `onChange` props (and that means we're no longer using local state to set those values)

jsudufour: An example of it implemented with the controller is in CarDamage.tsx (in the agent portal) on line 148. If you have any thoughts on making the changes you've mentioned above and still keep the component compatible with form interactions im keen to hear them!

Jmuskett: I *think* I have a working solution...  

so in the confirmation component in smores I've done this 

```
import React, { FC, FormEvent, ReactElement } from 'react'
import styled from 'styled-components'
import { Box } from '../Box'
import { Text } from '../Text'
import { RadioButton, FakeInput } from './RadioButtonStyled'

type ConfirmationProps = {
  checked?: boolean
  id: string
  error?: boolean
  errorMsg?: string
  label: string
  onBlur?: (e: FormEvent<HTMLInputElement>) => void
  sublabel?: string | ReactElement
  yesLabel?: string | ReactElement
  noLabel?: string | ReactElement
}

export const Confirmation: FC<ConfirmationProps> = ({
  checked,
  id,
  error = false,
  errorMsg = '',
  label,
  onBlur,
  sublabel,
  yesLabel = 'Yes',
  noLabel = 'No',
}: ConfirmationProps) => (
  <ConfirmationWrapper>
    <TextWrapper>
      <SectionHeadingText tag="h3">{label}</SectionHeadingText>
      {sublabel && (
        <Text tag="p" typo="base-small" color={theme.colors.grey8}>
          {sublabel}
        </Text>
      )}
    </TextWrapper>
    <RadioButtonGroupWrapper>
      <RadioButtonGroup>
        <RadioButtonWrapper checked={checked} error={error}>
          <RadioButton
            id={id}
            label={yesLabel}
            checked={checked}
            value={id}
            onBlur={onBlur}
          />
        </RadioButtonWrapper>
        <RadioButtonWrapper checked={!checked} error={error}>
          <RadioButton
            id={`${id}-1`}
            label={noLabel}
            checked={!checked}
            value={`${id}-1`}
            onBlur={onBlur}
          />
        </RadioButtonWrapper>
      </RadioButtonGroup>
      {error && <ErrorBox>{errorMsg}</ErrorBox>}
    </RadioButtonGroupWrapper>
  </ConfirmationWrapper>
)

const RadioButtonGroupWrapper = styled.div`
  display: flex;
  flex-direction: column;
`

const RadioButtonGroup = styled.div`
  display: flex;
  flex-direction: row;
`

const getColor = (checked?: boolean, error?: boolean) => {
  if (error) {
    return `2px solid ${theme.colors.red7}`
  } else if (checked) {
    return `2px solid ${theme.colors.blue7}`
  } else {
    return 'none'
  }
}

const RadioButtonWrapper = styled.div<FakeInput>`
  background-color: ${({ checked }: FakeInput) =>
    !checked && `${theme.colors.bg4}`};
  border: ${({ checked, error }: FakeInput) => getColor(checked, error)};
  margin: 0px 10px;
  width: 139px;
  display: flex;
  align-items: center;
  height: 56px;
  padding-left: 12px;
  border-radius: 5px;
  font-weight: bold;
  :hover {
    cursor: pointer;
  }
`

const ErrorBox = styled.span`
  margin-top: 7px;
  margin-left: 12px;
  font-size: 12px;
  color: ${theme.colors.red7};
`

const ConfirmationWrapper = styled(Box)`
  display: grid;
  grid-template-columns: 1fr 1fr;
  align-items: center;
`

const SectionHeadingText = styled(Text)`
  font-weight: bold;
`

const TextWrapper = styled.div`
  display: flex;
  flex-direction: column;
`

``` 

and then in CarDamage.tsx, from line 132 - we have this 

```
       <Controller
            name="damageDescription"
            control={control}
            defaultValue=""
            render={({ field: { onChange, onBlur, value } }) => (
              <TextareaStyled
                top
                id="damageDescription"
                label="damage description"
                placeholder="Enter damage description"
                onChange={onChange}
                onBlur={onBlur}
                value={value}
              />
            )}
          />
```

and that seems to work as expected. It seems like the `onChange` doesn't actually need to be consumed in the component itself, as it's just used by the controller locally to `CarDamage.tsx` 

let me know if this approach works/makes sense?

jsudufour: nice one, i think this makes sense! one change i would suggest is checking if `checked` is false explicitly so that when checked is undefined, false isn't selected. i'll ask Leo to have a look and provided everything is good will make the changes 👍

josianed:
deprioritized

PR #1571, opened 2021-11-17
Forward Ref and default props for Button and Chip component

fleonard:
## Screenshot / video

Screenshot 2021-11-17 at 12 36 00

## What does this do?

We can finally forward `ref` and any other default props (data and aria attributes for example) for `Button` and `Chip` components.
I've done these as a start since they are the most common use cases but if you think of any other that will need the same feel free to follow the same structure and improve them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [1.3.0] - 2021-11-17
+### Changed
+- Updated Button and Chip components to allow forwarding of ref and other default properties like data and aria attributes
+
 ## [1.2.24] - 2021-11-12
 ### Changed
 - Added time and date as types in text input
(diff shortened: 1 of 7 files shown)


ellajay (approved):

ellajay:
This is great!

trcywu (approved):

habin-isa (approved):
:shipit:

jsudufour:
out of curiosity, in what cases would you want to forward the ref on a button?

fleonard:
>out of curiosity, in what cases would you want to forward the ref on a button?

@jsudufour  Not something very common but you might need to manually set focus to a button after something happen. I mainly exposed ref just because you never know but the main scope of this PR was to expose all the default props that an HTML button has cause since now we couldn't add any data or a11y attribute to it

PR #1631, opened 2021-12-15
Introduce optional <optgroup> label to DropdownItem

joaocanaverde:
## What does this do?

Introduces optional [<optgroup>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) labels to dropdown items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [1.6.0] - 2021-12-15
+### Added
+- Added optional <optgroup> label to DropdownItem
 ## [1.5.1] - 2021-12-13
 ### Added
 - Added new icons
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -217,6 +217,7 @@
 ### Changed
 - Updated gap and styles on Row component
 
+[1.6.0]: https://github.com/marshmallow-insurance/smores-react/compare/v1.5.1...v1.6.0
 [1.5.1]: https://github.com/marshmallow-insurance/smores-react/compare/v1.5.0...v1.5.1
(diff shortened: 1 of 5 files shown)


src/Dropdown/Dropdown.stories.js
```diff
@@ -49,6 +87,14 @@ Default.args = {
   placeholder: 'Select Day',
 }
 
+export const DefaultWithGroups = Template.bind({})
+
+DefaultWithGroups.args = {
+  id: 'days',
+  list: daysWithGroups,
```
fleonard, line 94: Maybe better to keep the default one without groups?

fleonard (changes requested):
Looks good, can you please edit the CHANGELOG file so it will populate the release once you bump version and push tags?

src/Dropdown/Dropdown.stories.js
joaocanaverde: @fleonard We've got both still:

```javascript
export const Default = Template.bind({})

Default.args = {
  id: 'days',
  list: days,
  placeholder: 'Select Day',
}

export const DefaultWithGroups = Template.bind({})

DefaultWithGroups.args = {
  id: 'days',
  list: daysWithGroups,
  placeholder: 'Select Day',
}
```

fleonard: missed it

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## [1.6.0] - 2021-12-15
+### Added
+- Added optional <optgroup> label to DropdownItem
```
fleonard, line 3: you also need the link of the diff at the bottom of the file

joaocanaverde: @fleonard added

fleonard (approved):

PR #1663, opened 2022-01-06
Update brand colours

josianed:
## Screenshot / video

- Please check storybook to review all updated colours 

## What does this do?

- Migrate old colours to new brand colours as per [the migration guide](https://www.notion.so/getmarshmallow/Colour-migration-d2b20b1508fc4fa2b711aa7bf0a154b9)


src/theme.ts
```diff
@@ -1,79 +1,35 @@
...
-  blue3: '#94A0C6',
-  blue4: '#6F7AA0',
-  blue5: '#545E7E',
-  blue6: '#363F59',
-  blue7: '#112035',
-  blue8: '#05090E',
+  // Core brand
+  primary: '#FF4E75',
+  secondary: '#112035',
+  white: '#FFFFFF',
```
fleonard, line 5: I think we should also keep the black just in case even if it's not in the brand guide might still be useful

fleonard (commented):
Nice one 😍 
I would probably edit the changelog in here cause since it has breaking changes it makes sense to do a major bump and then we can see when we'll all be able to roll it out

src/Message/Message.tsx
```diff
@@ -40,7 +40,7 @@ export const Message: FC<MessageProps> = ({
       <Icon
         size={sizeSmall ? 24 : 32}
         render={type}
-        color={type === 'warning' ? 'pink8' : 'blue7'}
+        color={type === 'warning' ? 'primary' : 'secondary'}
```
fleonard, line 43: I've just realised that there's a small error with the migration guide. `pink8` has been used as an error color since it's basically a dark red so we should change it with `error` rather than `primary`

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## [2.0.0] - 2022-01-13
+### Added
+- Update brand colours
```
fleonard, line 3:
```suggestion
### Changed
- 💥 Migrated theme to new brand colours
```

fleonard (approved):

PR #1664, opened 2022-01-06
Fix - Make required prop in Dropdown component an optional prop on the component level

josianed:
## Screenshot / video

- No UI changes

## What does this do?

- Make required property a passed-in prop on the Dropdown component
- Defaulting to true preserves any existing functionality relying on this prop 
- Will disable the default required validation message when using the Dropdown with react-hook-forms and yup validation

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -49,6 +49,8 @@ type DefaultProps = {
   outlined?: boolean
   /** onBlur listener */
   onBlur?: (e: FormEvent<HTMLSelectElement>) => void
+  /** required item */
+  required?: boolean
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -83,6 +85,7 @@ export const Dropdown: FC<Props> = ({
   errorMsg = '',
   onInputChange,
   onBlur,
+  required = true,
```
fleonard, line 88: shouldn't this be false if it's optional?

fleonard (approved):

fleonard:
@josianed worth updating the changelog here if you are going to release straight away

PR #1665, opened 2022-01-06
Added AA inverted icon

fleonard:
## What does this do?

Added new inverted icon for AA


src/Icon/Container.jsx
```diff
@@ -9,6 +9,10 @@ export const Container = () => {
         <span>aa</span>
```
josianed, line 9: nit: do we need the curly brackets and explicit return?

josianed (approved):

src/Icon/Container.jsx
fleonard: no that's just used for the text. You can see how it works [here](https://marshmallow-insurance.github.io/smores-react/?path=/story/icon--collection) but it's essentialy a row of `iconRenderText      icon`

PR #1686, opened 2022-01-13
Minor hover updates with new colours

fleonard:
## What does this do?

- Added polished library to handle darken colours for button hover


habin-isa (approved):
:shipit:

PR #1694, opened 2022-01-19
docs: remove yarn from installation part

mickaelzhang:
Since we don't have a `yarn.lock` but have a `package-lock.json`, I guess we use npm only so we should probably remove `yarn` from the instruction?


fleonard:
@mickaelzhang considering that S'mores is a public library and everyone ca use it I think makes more sense to keep both option in there

mickaelzhang:
Ah, didn't consider this aspect 👍

PR #1698, opened 2022-01-20
Add color story

mickaelzhang:
## Screenshot / video

Screenshot 2022-01-20 at 16 50 27

## What does this do?

Added a story for the colors with the name, the hex and the color displayed as well

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,7 +23,7 @@
     "strict": ["error", "never"],
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "off",
-    "react/prop-types": "off"
+    "react/prop-types": "warn"
   },
   "overrides": [
     {
@@ -39,7 +39,8 @@
       ],
       "rules": {
         "@typescript-eslint/explicit-function-return-type": ["off"],
-        "@typescript-eslint/interface-name-prefix": ["off"]
+        "@typescript-eslint/interface-name-prefix": ["off"],
+        "react/prop-types": "off" // disables PropTypes for TS projects
       }
     }
   ]
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,8 @@
   "rules": {
     "strict": ["error", "never"],
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "off"
+    "react-hooks/exhaustive-deps": "off",
+    "react/prop-types": "off"
   },
   "overrides": [
     {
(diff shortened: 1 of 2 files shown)


.eslintrc
```diff
@@ -22,7 +22,8 @@
   "rules": {
     "strict": ["error", "never"],
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "off"
+    "react-hooks/exhaustive-deps": "off",
+    "react/prop-types": "off"
```
mickaelzhang, line 26: I moved this rules to be disabled everywhere but not sure if it was the best way to go, feedback welcomed
Other options could have been:
- Add prop-types, we would need to add the library tho as it wasn't installed
- Use TS in stories, I don't have context on why we are using JS in story

fleonard:
Looks super cool! Any chance you can make it looks like [this](https://www.figma.com/file/rnQu7stzZse0zw67V2tSy6/DS---Global%3A-Brand-assets-and-colours?node-id=330%3A0) so it will match Figma and it's easier to split colours by section?

.eslintrc
fleonard: No idea why we use js in stories. My preference would be to migrate them in Typescript if possible

mickaelzhang: Agree, it would be better if they were in TS

Make it a warning might be the better option for now 👍

mickaelzhang:
> Looks super cool! Any chance you can make it looks like this so it will match Figma and it's easier to split colours by section?

I'm not sure if you meant match it visually or separating in section like in Figma. I thought of both:
- **For the UI**
It would be harder to do as we would need to switch the text's color between black and white depending on the color
- **For the section**
I wasn't sure if was worth it to split it in section since I need to manually split them apart for each color. Meaning we would need to make some update if add/remove some color later

fleonard:
> * **For the UI**
>   It would be harder to do as we would need to switch the text's color between black and white depending on the color

Don't worry about the text colours. It was more about where to use each colour, you can potentially add a note to which text colour should be used based on that

> * **For the section**
>   I wasn't sure if was worth it to split it in section since I need to manually split them apart for each color. Meaning we would need to make some update if add/remove some color later, what do you think?

Feel free to do it manually - This colour update is been going around for ages and I think now we finally reached an agreement so I would be surprised if this will changed soon. Also we can easily tweak the story when it happens since we'll still need to update the theme anyway

fleonard:
@mickaelzhang I just noticed a small issue with the theme. Any chance you can swap `agentWarning` and `warning`?

AgentWarning should be orange (#FF8B00)
warning should be yellow (#FFC400)

mickaelzhang:
Nice catch, I'll just opened a PR here https://github.com/marshmallow-insurance/smores-react/pull/1699

The story is not merged yet and already useful! haha

mickaelzhang:
@fleonard Just made the update Leo, let me know if this was what you thought about!

fleonard (approved):
that's perfect 💯

habin-isa (approved):
This is amazing - great work 🔥🎉

PR #1713, opened 2022-01-26
Button border hover

fleonard:
## What does this do?

- fixed primary and tertiary buttons border colour on hover


Jmuskett (approved):
LGTM

PR #1721, opened 2022-01-28
Input component label enhancements

josianed:
## Screenshot / video

Screenshot 2022-01-28 at 17 17 37

## What does this do?

- Add props allowing to customize label size and color on input components
- The input can now be selected by label for tests even when the header's design is not the traditional small label

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -29,6 +29,16 @@ type DefaultProps = {
   placeholder?: string
   /** label displayed above the dropdown  */
   label?: string
+  /** label typography displayed above input  */
+  labelTypo?: string
+  /** label tag displayed above input  */
+  labelTag?: string
+  /** label color displayed above input  */
+  labelColor?: string
+  /** margin below label, when input is outlined */
+  outlinedLabelMargin?: string
+  /** margin below label, when input is not outlined */
+  labelMargin?: string
   /** used for label - input connection */
   name?: string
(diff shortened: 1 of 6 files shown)


mickaelzhang (commented):
As we discussed on Slack, let's check with the designer if there is a valid argument for adding a new variation

On the API update itself, instead of having a 1 to 1 mapping of the Text API, we could abstract it with the context of use? Something along the line of:
```tsx
<TextInput
  label="Username"
  // The prop name is TBD, something like variation/intent
  variation="contextOfUse"
  ...
/>
```

That way we could easily make an update if we see that the style change, it allows us to update the input style as well later if there is the need

Another solution could be to have a `Form` component that handle variation and pass the context to the form component's children:

```tsx
<Form variation="contextOfUse">
  // Every form components will take the form variation
  <TextInput ... />
  <RadioButton ... />
</Form>
```

src/Dropdown/Dropdown.tsx
```diff
@@ -29,6 +29,16 @@ type DefaultProps = {
   placeholder?: string
   /** label displayed above the dropdown  */
   label?: string
+  /** label typography displayed above input  */
+  labelTypo?: string
```
fleonard, line 33: Just to sums up what I think we should do here is to just keep this `labelTypo`, decide with design which other typo of the one we already have to use for bidder labels and remove all other props to keep it as basic as possible.

josianed:
deprioritized

PR #1741, opened 2022-02-07
TextInput - update input padding and cursor

mickaelzhang:
## Screenshot / video

## What does this do?

- Give the reviewers context

### DIFF
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -90,6 +90,7 @@ export const TextInput: FC<Props> = ({
         placeholder={placeholder}
         value={value}
         error={error}
+        outlined={outlined}
         autoComplete="off"
         onChange={(e: FormEvent<HTMLInputElement>) => {
           onChange && onChange(e.currentTarget.value)
@@ -120,16 +121,15 @@ interface IInput {
   error: boolean
   disabled?: boolean
   value?: string
-}
-
-interface IInputOutline extends IInput {
   outlined?: boolean
 }
 
-const Content = styled.div<IInputOutline>`
+const Content = styled.div<IInput>`
   border-bottom: 1px solid;
   border-color: ${({ error }) =>
     theme.colors[`${error ? 'error' : 'outline'}`]};
+  background-color: ${({ outlined }) =>
+    outlined ? 'transparent' : theme.colors['white']};
   display: flex;
   height: 32px;
 
@@ -144,7 +144,6 @@ const Content = styled.div<IInputOutline>`
     `
       border: 2px solid ${error ? theme.colors.error : theme.colors.outline};
       border-radius: 8px;
-      padding: 16px 12px;
       height: auto;
     `}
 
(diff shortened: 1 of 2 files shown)


src/TextInput/TextInput.tsx
```diff
@@ -158,12 +157,13 @@ const Content = styled.div<IInputOutline>`
 
 const Input = styled.input<IInput>`
   border: none;
+  background-color: transparent;
```
mickaelzhang, line 160: An `input` has a white background by default so I removed it as it was clashing with the border

src/TextInput/TextInput.tsx
```diff
@@ -158,12 +157,13 @@ const Content = styled.div<IInputOutline>`
...
   color: ${({ error }) => theme.colors[`${error ? 'error' : 'secondary'}`]};
   font-size: 16px;
   width: 100%;
   outline: none;
-  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
-  opacity: ${({ disabled }) => (disabled ? '0.5' : '1')};
+  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'initial')};
+  padding: ${({ outlined }) => (outlined ? '17px 14px' : '1px 2px')};
```
mickaelzhang, line 166: It was `16px 12px` but the input has some initial padding and I didn't go further to see if it would be possible to remove this additional padding

src/TextInput/TextInput.tsx
```diff
@@ -158,12 +157,13 @@ const Content = styled.div<IInputOutline>`
...
 const Input = styled.input<IInput>`
   border: none;
+  background-color: transparent;
   color: ${({ error }) => theme.colors[`${error ? 'error' : 'secondary'}`]};
   font-size: 16px;
   width: 100%;
   outline: none;
-  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
-  opacity: ${({ disabled }) => (disabled ? '0.5' : '1')};
+  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'initial')};
```
mickaelzhang, line 165: I'm not sure why it was pointer but I feel like user expect the initial cursor of an input for an input?

src/TextInput/TextInput.tsx
```diff
@@ -120,16 +121,15 @@ interface IInput {
...
   outlined?: boolean
 }
 
-const Content = styled.div<IInputOutline>`
+const Content = styled.div<IInput>`
   border-bottom: 1px solid;
   border-color: ${({ error }) =>
     theme.colors[`${error ? 'error' : 'outline'}`]};
+  background-color: ${({ outlined }) =>
+    outlined ? 'transparent' : theme.colors['white']};
```
mickaelzhang, line 132: Since I removed the background on the input itself, I put it back here
When it is the outline, I felt that not having one made sense

In most case, it wouldn't make a difference since we have mostly white background tho

src/TextInput/TextInput.tsx
```diff
@@ -158,12 +157,13 @@ const Content = styled.div<IInputOutline>`
 
 const Input = styled.input<IInput>`
   border: none;
+  background-color: transparent;
   color: ${({ error }) => theme.colors[`${error ? 'error' : 'secondary'}`]};
   font-size: 16px;
   width: 100%;
   outline: none;
-  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
-  opacity: ${({ disabled }) => (disabled ? '0.5' : '1')};
```
fleonard, line 166: any reason why the opacity is gone?

src/TextInput/TextInput.tsx
fleonard: Do you have a screenshot? does this maintain the 58px height of the input? Doesn't seem you have changed the height of it

src/TextInput/TextInput.tsx
mickaelzhang: It shouldn't be gone, probably by error

src/TextInput/TextInput.tsx
mickaelzhang: Yes, those value keep everything the same height as before so `58.5px`, not sure why there is the `.5` tho

fleonard (approved):

PR #1750, opened 2022-02-15
Top align multiline labes in checkboxes

fleonard:
## Screenshot / video

Screenshot 2022-02-14 at 17 56 19

## What does this do?

- Give the reviewers context

### DIFF
--- a/src/CheckBoxGroup/CheckBoxGroup.tsx
+++ b/src/CheckBoxGroup/CheckBoxGroup.tsx
@@ -6,7 +6,6 @@ import { CheckBox } from './CheckBox'
 type Item = {
   id: number
   label: string
-  value: string
   checked: boolean
 }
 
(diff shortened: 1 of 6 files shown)


src/CheckBoxGroup/CheckBoxGroup.tsx
```diff
@@ -6,7 +6,6 @@ import { CheckBox } from './CheckBox'
 type Item = {
   id: number
   label: string
-  value: string
```
mickaelzhang, line 9: Why are we removing the value?

mickaelzhang (approved):

src/CheckBoxGroup/CheckBoxGroup.tsx
fleonard: it's been there since the beginning but it's never used or passed down the checkbox. If you think make more sense I can pass it down to the input itself?

mickaelzhang: Just wanted to make sure but it makes sense to remove it if it's not used 👍

PR #1756, opened 2022-02-17
Adds included and excluded icons

githab666:
## Screenshots
Screenshot 2022-02-17 at 10 56 03
Excluded
Screenshot 2022-02-17 at 10 55 56
Included

## What does this do?
- Adds `included` and `excluded` to `Icon` component
- Replaces raw svg with corresponding `Icon` components for the `IconWrapper` component


githab666:
Will bump tags once approved

src/Icon/Icon.tsx
```diff
@@ -1268,6 +1268,20 @@ export const Icon: FC<Props> = ({
         />
       </svg>
     )}
+    {render === 'excluded' && (
+      <svg width="100%" height="100%" viewBox="0 0 32 32" fill="none">
+        <path
+          fillRule="evenodd"
+          clipRule="evenodd"
+          d="M16 32C24.8366 32 32 24.8366 32 16C32 7.16344 24.8366 0 16 0C7.16344 0 0 7.16344 0 16C0 24.8366 7.16344 32 16 32Z"
+          fill={theme.colors.error}
```
fleonard, line 1277: can we make these use the colour prop instead? better keep them generic even tho they are used with only red and green colour just in case there's gonna be a different use case where we need them with the default secondary background colour. Also the render name should match what designers uses in Figma so can you please ask Luke to add these two in the DS Global Iconography file and name them the same? otherwise it all ends up out of sync

githab666: Ok cool!

fleonard (approved):

PR #1757, opened 2022-02-17
Update Icon so that it never shrink

mickaelzhang:
## Screenshot / video

**BEFORE**
https://user-images.githubusercontent.com/14129033/154464227-d1dabc1e-840b-413b-89b9-76a91d5cd483.mp4

**AFTER**
https://user-images.githubusercontent.com/14129033/154464296-9a6ae186-6f4a-445a-a1cb-7f4687eaa1a3.mp4

## What does this do?

In some case, the `Icon` could shrink even when the size was set. I believe this isn't the expected behaviour, where we write:
```jsx
<Icon render="icon-name" size={24} />
```
We expect the icon to stay at 24px and not decrease or increase in any context, except if explicitly declared by changing the size in the code


src/Icon/Icon.tsx
```diff
@@ -4090,8 +4090,8 @@ const Container = styled.span<IIcon>(
     display: flex;
     align-items: center;
     justify-content: center;
-    width: 100%;
-    max-width: ${size}px;
```
mickaelzhang, line 4094: I'm not sure why we set the width this way?

src/Message/Message.tsx
```diff
@@ -61,7 +61,6 @@ interface IWrapper {
 }
 
 const IconWrapper = styled(Box)<IIconWrapper>`
-  flex-shrink: 0;
```
mickaelzhang, line 64: This isn't necessary anymore

fleonard (approved):

PR #1763, opened 2022-02-22
Ft pagination handle smaller page numbers

CallumMarriage:
## Screenshot / video

## What does this do?

- Fixes a bug with the pagination component where the last page is not updated if moving between a list that has more than 1 page and a list that has a single page.
- Fixes a bug with the pagination component where the "pages icons" are not updated if moving between a list that has more than 2 pages and a list that has two pages or less.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [2.1.5] - 2022-02-23
+### Changed
+- Fix Pagination bug for small number of pages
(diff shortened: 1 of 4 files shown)


fleonard (commented):
looks good - do you mind updating the CHANGELOG file following the same structure as the other releases notes? You just need to add the new version + date and description of what's changed at the top and the link of the diff version at the bottom of the file

fleonard (approved):

PR #1773, opened 2022-02-28
fix button layout shifting when a icon is present or when loading

mickaelzhang:
## Screenshot / video

**BEFORE**

https://user-images.githubusercontent.com/14129033/156031389-6dd916a6-1b3f-4dfb-be19-829a7246de38.mov

The button has a height of 52px but when an icon is present, it goes to 60px ([link to the storybook](https://marshmallow-insurance.github.io/smores-react/?path=/story/button--icon)).
When going to a loading state, the button size changed as well

**AFTER**

https://user-images.githubusercontent.com/14129033/156031419-aaf92b55-482f-49d8-8d6f-5ccb32ff36a9.mov

The button stays at 52px height and there is no size change during loading state change

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "test": "jest --silent",
     "snyk-protect": "snyk protect",
     "prepare": "npm run snyk-protect && husky install",
-    "pre-commit-lint": "npm run check-types && lint-staged"
+    "pre-commit-lint": "npm run check-types && lint-staged",
+    "lint": "eslint"
   },
   "devDependencies": {
     "@babel/core": "^7.13.15",
@@ -51,7 +52,7 @@
     "eslint": "^7.9.0",
     "eslint-config-prettier": "^8.2.0",
     "eslint-plugin-jest": "^26.0.0",
-    "eslint-plugin-react": "^7.23.2",
+    "eslint-plugin-react": "^7.29.2",
     "eslint-plugin-react-hooks": "^4.1.2",
     "husky": "^7.0.0",
     "jest": "^26.4.2",
--- a/package.json
+++ b/package.json
@@ -25,8 +25,7 @@
     "test": "jest --silent",
     "snyk-protect": "snyk protect",
     "prepare": "npm run snyk-protect && husky install",
-    "pre-commit-lint": "npm run check-types && lint-staged",
-    "lint": "eslint"
+    "pre-commit-lint": "npm run check-types && lint-staged"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
     "eslint": "^7.9.0",
     "eslint-config-prettier": "^8.2.0",
     "eslint-plugin-jest": "^26.0.0",
-    "eslint-plugin-react": "^7.23.2",
+    "eslint-plugin-react": "^7.29.2",
```
mickaelzhang, line 54: I had an issue with the `jsx-key` rule, it seems that there is an issue with this specific version and upgrading fix the issue

src/Button/Button.tsx
```diff
@@ -72,46 +72,9 @@ export const Button: FC<ButtonProps> = forwardRef<
...
-            </>
-          )}
-        </Container>
-      ) : (
+  ) => {
+    if (!primary && !secondary && !tertiary) {
+      return (
         <LegacyButton
           id={id}
           className={className}
```
mickaelzhang, line 80: Moved the legacy button to an early return to remove a bit of noise from the return element

src/Button/Button.tsx
```diff
@@ -126,33 +89,63 @@ export const Button: FC<ButtonProps> = forwardRef<
...
+        <ContentContainer icon={icon} loading={loading}>
+          {icon && (
+            <IconContainer
+              render={icon}
+              size={24}
+              color={primary ? 'white' : 'secondary'}
+            />
+          )}
+          <ChildrenContainer>{children}</ChildrenContainer>
+        </ContentContainer>
```
mickaelzhang, line 131: The content is now never removed to avoid the size to change during state change, instead we make the content transparent when in a loading state

src/Button/Button.tsx
```diff
@@ -72,46 +72,9 @@ export const Button: FC<ButtonProps> = forwardRef<
       ...props
     },
     ref,
-  ) => (
-    <div>
```
mickaelzhang, line 76: I removed this unnecessary `div` as it kind of mess with some styling

src/Button/Button.tsx
fleonard: I'm hoping at some point to get rid of that legacy button - I think we are in a good place but still not 100% sure

fleonard (approved):
👍🏻

PR #1861, opened 2022-04-14
Migrate stories to TS

mickaelzhang:
## What does this do?

- Moved all stories and file to TS
- Only js file are now config one


src/Accordion/Accordion.stories.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { Accordion } from './Accordion'
+import { Accordion, AccordionProps } from './Accordion'
 
 export default {
   title: 'Accordion',
   component: Accordion,
 }
 
-const Template = (args) => (
+const Template = (args: AccordionProps) => (
```
Jmuskett, line 9: super nit, but shouldn't this be

```
props: AccordionProps
```

OR

```
args: accordionArgs
```
?

src/Box/Box.stories.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { Box } from './Box'
+import { Box, BoxProps } from './Box'
 
 export default {
   title: 'Box',
   component: Box,
 }
 
-const Template = (args) => <Box {...args}>Some lovely content.</Box>
+const Template = (args: BoxProps) => <Box {...args}>Some lovely content.</Box>
```
Jmuskett, line 9: same comment as above, I think let's go for consistent naming on these?

Jmuskett (changes requested):
looks good on the whole, just a question about where we are using `args` but typing it as `props` - I know it's a real nit but I think we should be consistent with the naming, either props or args

src/Accordion/Accordion.stories.tsx
mickaelzhang: I didn't think about changing anything in the stories, except when it was related to typing
The typing is related to props so it should definitely be `AccordionProps`
For the other one, I don't mind either way, I'll update it to `props` 👍

PR #1869, opened 2022-04-21
Add scrolling on desktop when content is too long

mickaelzhang:
## Screenshot / video

| BEFORE | AFTER |
| ------------- | ------------- |
| Screenshot 2022-04-21 at 12 26 09 | Screenshot 2022-04-21 at 12 19 30  |

## What does this do?

- Add scrolling when the content is too long on desktop
- Updated scrolling property from `scroll` to `auto` on mobile, there should be no difference but this way, the scrolling bar isn't shown except when there is scrollable content

### DIFF
--- a/src/Modal/Modal.stories.tsx
+++ b/src/Modal/Modal.stories.tsx
@@ -3,8 +3,25 @@ import { Button } from '../Button'
 import { Box } from '../Box'
 import { Modal, ModalProps } from './Modal'
 
+const placeholderText =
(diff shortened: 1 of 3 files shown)


fleonard:
@mickaelzhang this will potentially break quite a lot of existing modals since as far as I remember when the content was longer than the modal we manually added wrappers so you might end up having multiple divs with scrolling behaviour

mickaelzhang:
@fleonard Mmm, you're right 🤔  We should definitely take that into account

Fixing this on S'mores seems to be the right approach for the long term so it might still be worth it? I'm sure we have this bug in a few places

fleonard:
@mickaelzhang oh yeah - we should definitely fix it in Smores. I'm just not sure how we can easily check where we need to manually fix the modals after this is published

mickaelzhang:
@fleonard Usage of Modal should be fairly small so checking manually should be possible?
I think it would be safe to just remove any container with a height + overflow set

fleonard (approved):

PR #1885, opened 2022-04-27
Grw 1483 - Optional label prop for Confirmation

githab666:
## Screenshot
Screenshot 2022-04-27 at 13 51 45

## What does this do?
- Adds optional `label` prop for `Confirmation` component
- Updates `CHANGELOG` - adds missing version from previous version too
- Adds `With no label` definition in Storybook

### DIFF
--- a/src/ConfirmationRadioButtons/Confirmation.tsx
+++ b/src/ConfirmationRadioButtons/Confirmation.tsx
@@ -11,7 +11,7 @@ export type ConfirmationProps = {
   id: string
   error?: boolean
   errorMsg?: string
-  label: string
+  label?: string
   onBlur?: (e: FormEvent<HTMLInputElement>) => void
   sublabel?: string | ReactElement
   yesLabel?: string | ReactElement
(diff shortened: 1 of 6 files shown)


fleonard:
@habin-isa What's the use case of having a YES/NO pair of radio buttons without a label?

src/ConfirmationRadioButtons/Confirmation.tsx
```diff
@@ -32,14 +32,17 @@ export const Confirmation: FC<ConfirmationProps> = ({
...
-      <TextWrapper>
-        <SectionHeadingText tag="h3">{label}</SectionHeadingText>
-        {sublabel && (
-          <Text tag="p" typo="base-small" color={theme.colors.subtext}>
-            {sublabel}
-          </Text>
-        )}
-      </TextWrapper>
+      {label ||
+        (sublabel && (
```
fleonard, line 36: should this be `{(label || sublabel) && (...`

mickaelzhang (commented):
Would be great if you could provide the context and in which case we don't want the label!

In terms of accessibility, a label should always be required
Instead of making it optional, we should hide it on screen but still show it for screenreaders

githab666:
> Would be great if you could provide the context and in which case we don't want the label!
> 
> In terms of accessibility, a label should always be required Instead of making it optional, we should hide it on screen but still show it for screenreaders

We need it for additional questions for students work i'm doing:
Screenshot 2022-04-27 at 17 35 01

fleonard:
@githab666 why can't we add a prop to change the layout to column instead of row instead? That to me looks exactly like the component we currently have but with the label on top

mickaelzhang:
Agree with Leo, it's a bit weird that the Confirmation component is the only one with a horizontal layout tho 🤔
Not sure if anyone has context on the why

githab666:
Paired with @mickaelzhang and due to the label being on the page for other questions in sign up have hidden it. We need a wider conversation about adding labels to forms in the next Frontend Guild. Also added an empty div when hidden to stop screen reader's being confused.

githab666:
Screenshot 2022-04-28 at 12 31 51
In this use case the titles for form fields are on page level not S'mores for sign up additional questions page

fleonard (approved):
@mickaelzhang thie component was done specifically for FNOL and there questions have an horizontal layout. 
I suppose we should ideally rework this as some sort of radio group with different style of the existing one and have an option to add the label either on top or next to it.

PR #1898, opened 2022-05-04
Add focus outline on CheckBox, CheckBoxGroup, RadioButton and Toggle

mickaelzhang:
## What does this do?

Add focus outline for a few components:
- CheckBox
- CheckBoxGroup
- RadioButton
- Toggle


src/CheckBoxGroup/CheckBox.tsx
```diff
@@ -1,83 +0,0 @@
...
-type Props = {
-  id: string
-  checked: boolean
-  children: ReactNode
-  toggle: () => void
-}
-
-export const CheckBox: FC<Props> = ({ id, checked, children, toggle }) => (
-  <BoxContainer id={id}>
-    <Text tag="span" typo="base">
```
mickaelzhang, line 16: This component was only used for `CheckBoxGroup`, I've deleted to use the `CheckBox` component instead as it was a copy paste of the Checkbox component

src/CheckBoxGroup/CheckBoxGroup.tsx
```diff
@@ -30,6 +30,7 @@ export const CheckBoxGroup: FC<Props> = ({ list, onToggle }) => (
 )
 
 const Container = styled.div`
-  display: grid;
-  grid-row-gap: 16px;
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
```
mickaelzhang, line 35: Not sure why `grid` was used, it's a 2D layout here so I've used `flex` sintead

src/CheckBox/CheckBox.tsx
```diff
@@ -22,12 +23,12 @@ export const CheckBox: FC<CheckBoxProps> = ({
...
 }) => (
   <>
-    <BoxContainer id={id}>
+    <BoxContainer>
       <Text tag="span" typo="base" color={error ? 'red7' : 'secondary'}>
         {children}
       </Text>
 
-      <input type="checkbox" checked={checked} onChange={toggle} />
+      <input id={id} type="checkbox" checked={checked} onChange={toggle} />
```
fleonard, line 31: worth checking that this won't break any test? If some checkbox test where done looking for `id` and then finding the input inside moving it will beak them

githab666 (commented):
Looks good but check on input id

src/CheckBox/CheckBox.tsx
mickaelzhang: I've reverted back the change

fleonard: Yeah probably safest to keep the I'd where it is for now but we should move it to the input cause it's the right place

fleonard (approved):

PR #1899, opened 2022-05-04
add link component

mickaelzhang:
[Related RFC](https://www.notion.so/getmarshmallow/Link-RFC-934e63db272d414f98313c25c7f96321)

Waiting a few days just in case anyone has a comment but ready to be reviewed

### DIFF
--- a/src/Link/Link.tsx
+++ b/src/Link/Link.tsx
@@ -1,0 +1,98 @@
+import React, { FC, MouseEventHandler, ReactNode } from 'react'
+import styled, { css } from 'styled-components'
+import { darken } from 'polished'
+
+import { focusOutline } from '../utils/focusOutline'
+import { theme } from '../theme'
+import { Icon } from '../Icon'
+
+export type LinkProps = {
+  className?: string
+  href: string
+  onClick?: MouseEventHandler
+  openInNewTab?: boolean
+  children?: ReactNode
+}
+
+export const Link: FC<LinkProps> = ({
+  className,
+  href,
+  onClick,
+  openInNewTab,
+  children,
+}) => {
+  return (
+    <LinkWrapper
+      href={href}
+      className={className}
+      onClick={onClick}
+      {...(openInNewTab && {
+        rel: 'noopener noreferrer',
+        target: '_blank',
+      })}
+    >
+      {children}
+      {openInNewTab && (
+        <StyledIcon render="new-window" size={12} color="primary" />
+      )}
+    </LinkWrapper>
+  )
+}
+
+const LinkWrapper = styled.a`
+  ${focusOutline}
+  display: inline-flex;
+  flex-direction: row;
+
+  font-size: 14px;
+  font-weight: ${theme.font.weight.medium};
+  line-height: 14px;
+  text-decoration: underline;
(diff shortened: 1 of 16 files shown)


src/Link/index.ts
```diff
@@ -0,0 +1 @@
+export { Link, LinkProps } from './Link'
```
mickaelzhang, line 1: We don't want to import everything, `linkStyleOverride` is internal but exported from `Link.tsx`

src/Tag/__tests__/Tag.js
```diff
@@ -1,12 +1,12 @@
...
+import { Tag } from '../Tag'
 
 test('renders', () => {
-  const {container} = render(
-    <Tag label="Cancelled - Refunded Manually" color="red7" />,
-  );
-  expect(container.firstChild).toMatchSnapshot();
-});
+  const { container } = render(
+    <Tag label="Cancelled - Refunded Manually" color="error" />,
```
mickaelzhang, line 9: Seems like this was incorrect but never picked up by the test

src/Text/Text.tsx
```diff
@@ -219,11 +220,12 @@ const Container = styled.p<IText>(
...
 
-  margin: 0;
+    margin: 0;
     padding: 0;
-    color: ${theme.colors[color]};
     text-align: ${align};
     cursor: ${cursor};
     letter-spacing: 0.15px;
+    color: ${theme.colors[color]};
+    ${linkStyleOverride({ color: theme.colors[color] })}
```
mickaelzhang, line 229: This work to override the colour, for the hover as well
It doesn't work for the active state and the font size, I don't think it's worth spending more effort into this as the `Text` component will be revamped in the future and we can find a more scalable way to handle it at this point

src/Link/Link.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { focusOutline } from '../utils/focusOutline'
+import { theme } from '../theme'
+import { Icon } from '../Icon'
+
+export type LinkProps = {
+  className?: string
+  href: string
+  onClick?: MouseEventHandler
+  openInNewTab?: boolean
+  children?: ReactNode
```
fleonard, line 14: Didn't we also had a download prop?

mickaelzhang: Yes it was, I forgot to bring it up but from what I read `download` is useless for cross-browser file/content
It works only when the URL is from the same domain, knowing this, I wasn't sure if it made sense to have it in the component?

fleonard: yeah but in our case the download should only be applied to pdf that are usually uploaded to `assets.marshmallow.com` so we shouldn't have any issue with CORS.
If we wanna drop the `download` attribute then I think we should change the way `openInNewTab` and the icon works cause in home we will need to open the PDFs in a new tab but the icon of the link is different

This is what has been designed to open/download the IPID docs in home 
Screenshot 2022-05-05 at 08 30 41

mickaelzhang: I thought most of our files were uploaded in a S3 bucket, my bad! I'll add it back

fleonard (approved):

PR #1924, opened 2022-05-19
fix test and remove uuid

mickaelzhang:
## Screenshot / video

## What does this do?

- Give the reviewers context

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.9",
     "@types/styled-components": "^5.1.3",
-    "@types/uuid": "^8.3.0",
     "@typescript-eslint/eslint-plugin": "^4.3.0",
     "@typescript-eslint/parser": "^4.3.0",
     "babel-eslint": "^10.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -56,6 +55,7 @@
     "eslint-plugin-react-hooks": "^4.1.2",
     "husky": "^8.0.1",
     "jest": "^28.0.3",
+    "jest-environment-jsdom": "^28.1.0",
```
mickaelzhang, line 58: Screenshot 2022-05-19 at 16 59 55

You need to install `jest-environment-jsdom` since `jest` v28 apparently

src/Pagination/Pagination.tsx
```diff
@@ -103,7 +102,7 @@ export const Pagination: FC<PaginationProps> = ({
         return (
           <PageBox
             active={activePage === i}
-            key={UUID()}
+            key={i}
```
mickaelzhang, line 105: Indexing using the page number will work better in this case as the page number is unique and doesn't change between render

It also allow us to remove a dependency which isn't needed

fleonard (approved):

PR #1925, opened 2022-05-19
Add ref properly to input components

mickaelzhang:
## Screenshot / video

## What does this do?

- Give the reviewers context

### DIFF
--- a/src/SearchInput/SearchInput.tsx
+++ b/src/SearchInput/SearchInput.tsx
@@ -195,6 +195,8 @@ const Input = styled.input<ISearchInput>`
 
   &::placeholder {
     color: ${theme.colors.subtext};
+  }
+
   ${({ outlined }) =>
     outlined &&
     `
--- a/src/Textarea/Textarea.tsx
+++ b/src/Textarea/Textarea.tsx
@@ -1,4 +1,4 @@
-import React, { FC, FormEvent, RefObject } from 'react'
+import React, { FormEvent, ForwardedRef, forwardRef } from 'react'
 import styled from 'styled-components'
 import { darken } from 'polished'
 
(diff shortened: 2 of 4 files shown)


src/Textarea/Textarea.tsx
```diff
@@ -12,8 +12,6 @@ type BaseProps = {
   id: string
   /** className attribute to apply classses from props */
   className?: string
-  /** ref attribute for input */
-  ref?: RefObject<HTMLTextAreaElement>
```
mickaelzhang, line 16: Adding it to the props shouldn't do anything, I've wrapper the component in `forwardRef` instead

src/SearchInput/SearchInput.tsx
```diff
@@ -195,6 +195,8 @@ const Input = styled.input<ISearchInput>`
 
   &::placeholder {
     color: ${theme.colors.subtext};
+  }
+
```
mickaelzhang, line 199: It seems to have been missing since the color migration

fleonard:
@mickaelzhang I'm wondering if on top of forwarding ref we should also forward all the other missing props in order to be able to use data attributes and such? I've already done it in the Button and Chip components.
Just not 100% sure what exactly we should fw since the form elements are nested

mickaelzhang:
@fleonard I think we should but it's definitely trickier as you said, there will probably be the need to split the props around either the container component and the input 🤔 

I don't think it's related to this PR so maybe we can take a look at it separately?

fleonard:
> @fleonard I think we should but it's definitely trickier as you said, there will probably be the need to split the props around either the container component and the input 🤔
> 
> I don't think it's related to this PR so maybe we can take a look at it separately?

@mickaelzhang yeah definitely not part of this PR. Was just a random question if you had some sort of idea already

fleonard (approved):

PR #1934, opened 2022-05-23
Fix Datepicker range display

mickaelzhang:
## Screenshot / video

## What does this do?

- Give the reviewers context

### DIFF
--- a/src/Datepicker/Datepicker.tsx
+++ b/src/Datepicker/Datepicker.tsx
@@ -3,17 +3,14 @@ import styled from 'styled-components'
 import {
   format,
   addDays,
-  addMonths,
   getMonth,
   getYear,
+  eachMonthOfInterval,
   isSameDay,
-  isSameMonth,
   isWithinInterval,
-  startOfMonth,
   getDaysInMonth,
   isToday,
   isWeekend,
-  getISODay,
 } from 'date-fns'
 
 import { Box } from '../Box'
@@ -22,7 +19,6 @@ import { Icon } from '../Icon'
 import { theme } from '../theme'
 
 import { DatesList } from './DatesList'
-import { Month } from './types'
 
 export type DatepickerProps = {
   disableWeekend?: boolean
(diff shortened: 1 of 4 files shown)


src/Datepicker/Datepicker.tsx
```diff
@@ -131,31 +91,29 @@ export const Datepicker: FC<DatepickerProps> = ({
         direction="row"
       >
         <Circle
-          disabled={listDays[activeMonth].monthName === listDays[0].monthName}
-          onClick={() => setActiveMonth(activeMonth - 1)}
+          disabled={activeMonthIndex === 0}
+          onClick={() => setActiveMonth(activeMonthIndex - 1)}
```
mickaelzhang, line 95: We were comparing string before but we can compare the index directly instead which would achieve the same

src/Datepicker/Datepicker.tsx
```diff
@@ -84,44 +74,14 @@ export const Datepicker: FC<DatepickerProps> = ({
...
-        endDay,
-      )
-
-      availableDays = [thisMonth, nextMonth]
-    } else {
-      availableDays = [generateDaysForMonth(today, endDay)]
-    }
-
-    setListDays(availableDays)
-  }
```
mickaelzhang, line 119: We were calculating all month before, I switched to only generate the days of the current displayed month

src/Datepicker/Datepicker.tsx
```diff
@@ -22,7 +19,18 @@ import { Icon } from '../Icon'
...
+  const monthList = eachMonthOfInterval({
+    start: startDate,
+    end: endDate,
+  })
+
+  return monthList.map((monthDate) => ({
+    date: monthDate,
+    label: format(monthDate, 'MMMM'),
+  }))
+}
```
mickaelzhang, line 33: Only the months are generated in advance, although there is probably a case to be made for this to be generated progressively as well
For case where there is no range for example, although it doesn't seems it's possible with the component today

src/Datepicker/DatesList.tsx
```diff
@@ -1,29 +1,46 @@
...
+  items,
+  handleDateSelect,
+  firstDayShift,
+}) => {
+  return (
+    <Container>
+      {firstDayShift &&
+        Array(getBlankDaysCount(items[0].date))
+          .fill(null)
+          .map((_, index) => <ListItem key={`blankItem-${index}`} disabled />)}
```
mickaelzhang, line 30: I moved the blank day (calendar display mode) inside the list, I don't think it made sense to be inside the day generation function as it's not related to the data itself, only presentational

src/Datepicker/Datepicker.tsx
```diff
@@ -39,42 +47,24 @@ export const Datepicker: FC<DatepickerProps> = ({
   fromDate = new Date(),
```
fleonard, line 47: Were you planning to solve the timezone issue in signup? I'm wondering if might be better to set it by default to generate UK time only start dates

fleonard (approved):
LGTM, just a minor comment about timezones

src/Datepicker/Datepicker.tsx
mickaelzhang: Yeah, planning to fix the time zone issue on a separate PR! Will open the PR this morning

PR #1938, opened 2022-05-24
New Support Message component

githab666:
## Screenshot
Screenshot 2022-06-06 at 11 34 30

## What does this do?
- Adds new Support message component
- Updates CHANGELOG


src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+  className,
+  children,
+  type = 'info',
+  hasBorder,
+  borderColor,
+  title,
+  hasBackground = true,
+}) => (
+  <Wrapper
+    className={className}
```
fleonard, line 30: since wrapper is a div instead of manually adding class name shouldn't we just extend the type like this and spread all html div props?

````
SupportMessageProps & HTMLAttributes<HTMLDivElement> = ({
   // existing props
   ... props
   })
```

and then in the wrapper you can do

`<Wrapper //existing props {...props}>`

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+    </IconWrapper>
+    <Box flex direction="column">
+      <Title>{title}</Title>
+      {children}
+    </Box>
+  </Wrapper>
+)
+
+interface IWrapper {
+  type: 'info' | 'alert' | 'warning'
```
fleonard, line 57: you should be able to get the type from `SupportMessageProps` so it always match

src/SupportMessage/__tests__/SupportMessage.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import 'jest-styled-components'
+
+import { SupportMessage } from '../SupportMessage'
+
+it('default renders correctly', () => {
```
fleonard, line 8: worth wrapping all of these in a describe?

src/SupportMessage/SupportMessage.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default {
+  title: 'SupportMessage',
+  component: SupportMessage,
+}
+
+const Template = (props: SupportMessageProps) => (
+  <SupportMessage {...props}>Support message for customer</SupportMessage>
+)
+
+export const Default = Template.bind({})
```
fleonard, line 14: you need to bing the args of the default template if you want to make it interactive

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+          type === 'warning'
+            ? 'error'
+            : type === 'alert'
+            ? 'agentWarning'
+            : 'secondary'
+        }
+      />
+    </IconWrapper>
+    <Box flex direction="column">
+      <Title>{title}</Title>
```
fleonard, line 50: title is optional so this should be conditional to avoid rendering an empty `h3`

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Icon } from '../Icon'
+import { theme } from '../theme'
+
+type BorderProps =
+  | { hasBorder?: false; borderColor?: never }
+  | { hasBorder: true; borderColor: string }
+
+export type SupportMessageProps = {
+  className?: string
+  children: ReactNode
```
fleonard, line 14: Based on the current design shouldn't this be just a string? Having it as a ReactNode means we can render as a child what we want

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+const Wrapper = styled.div<IWrapper>(
+  ({ type, hasBorder, borderColor, hasBackground }) => css`
+    align-items: center;
+    ${hasBackground &&
+    `background-color: ${
+      type === 'info'
+        ? theme.colors.background
+        : type === 'alert'
+        ? theme.colors.bgSecondary
+        : '#FBEAEA'
```
fleonard, line 76: we shouldn't use colours that are not in our palette. I think this has to bee some sort of lighter version of `error`

src/SupportMessage/__tests__/SupportMessage.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import 'jest-styled-components'
+
+import { SupportMessage } from '../SupportMessage'
+
+describe('SupportMessage component', () => {
+  it('default renders correctly', () => {
```
fleonard, line 9: I think all these tests are basically doing exactly the same thing so not sure how worth they are?
Probably better to start avoiding using snapshots for new components and test them properly with testing library?

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+const Wrapper = styled.div<IWrapper>(
+  ({ type, hasBorder, borderColor, hasBackground }) => css`
+    align-items: center;
+    ${hasBackground &&
+    `background-color: ${
+      type === 'info'
+        ? theme.colors.background
+        : type === 'alert'
+        ? theme.colors.bgSecondary
+        : lighten(0.45, theme.colors.error)
```
mickaelzhang, line 75: To make it more readable, you could create an object with the type as key maybe?

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+    <IconWrapper>
+      <Icon
+        size={24}
+        render={type}
+        color={
+          type === 'warning'
+            ? 'error'
+            : type === 'alert'
+            ? 'agentWarning'
+            : 'secondary'
```
mickaelzhang, line 44: Same here, create an object

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { lighten } from 'polished'
+
+import { Box } from '../Box'
+import { Icon } from '../Icon'
+import { theme } from '../theme'
+
+export type SupportMessageProps = {
+  className?: string
+  children: string
+  type: 'info' | 'alert' | 'warning'
```
mickaelzhang, line 12: I'm guessing we need to align first with design if we talk about `type` or something else?

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+  title?: string
+  hasBackground?: boolean
+  hasBorder?: boolean
+  borderColor?: string
```
mickaelzhang, line 16: Those seems too low level in term of customisation
I think we need to clarify a few points with design before finding a good solution on this part

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+import styled, { css } from 'styled-components'
+import { lighten } from 'polished'
+
+import { Box } from '../Box'
+import { Icon } from '../Icon'
+import { theme } from '../theme'
+
+export type SupportMessageProps = {
+  className?: string
+  children: string
```
mickaelzhang, line 11: Instead of using `children`, should we use a named prop? Like `description`? It would make it more explicit

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+  hasBorder?: boolean
+  borderColor?: string
+  hasBackground?: boolean
+}
+
+const IconWrapper = styled(Box)`
+  align-self: center;
+`
+
+const Wrapper = styled.div<IWrapper>(
```
mickaelzhang, line 66: I think the wrapper should be `width:100%` by default?

mickaelzhang (changes requested):

src/SupportMessage/SupportMessage.tsx
fleonard: I mean, we don't really care much on how design call them since in Figma you won't really see a name, you just infer from the look of the component which type/variation we need to have.

src/SupportMessage/SupportMessage.tsx
fleonard: you were thinking to have an extra type for the bordered one instead?

src/SupportMessage/SupportMessage.tsx
fleonard: Agree, we should call this `description`

src/SupportMessage/SupportMessage.tsx
mickaelzhang: We can see it on Figma as well, it's currently named state

Screenshot 2022-05-26 at 09 43 52

src/SupportMessage/SupportMessage.tsx
mickaelzhang: This could be a potential solution yes
Not sure about the best way to go but it would be better as it would pair the background, border and border color together instead of having it separated like that

src/SupportMessage/SupportMessage.tsx
githab666: I matched them to icons used otherwise naming doesn't make sense and too disruptive to change Icon names directly

fleonard: I think Mickael was talking about the prop name not the values. I guess we can go with `state` if that's what's used in Figma

src/SupportMessage/SupportMessage.tsx
fleonard: To pair everything together we need to rely on the whole state prop and just style it according to it. I think it's easily doable and it's definitely less customisable and more close to Figma designs.
The only problem is that design will have to change the info state cause currently has both a bordered and non bordered version

src/SupportMessage/SupportMessage.tsx
githab666: AH! I see i misunderstood. Ok i can change to state

mickaelzhang: From what I've seen, they use state for everything currently but I think we can talk with design so they use `type` or something more explicit instead
`state` is too general as a term

fleonard (approved):
LGTM

package.json
```diff
@@ -44,6 +44,8 @@
     "@types/node": "^17.0.35",
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.9",
+    "@types/testing-library__jest-dom": "^5.14.3",
+    "@types/testing-library__react": "^10.2.0",
```
mickaelzhang, line 48: This shouldn't be needed if you rebase your PR?

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,97 @@
...
+    icon: 'warning',
+  },
+}
+
+export type SupportMessageProps = {
+  className?: string
+  description: string
+  type: 'info' | 'info-outline' | 'alert' | 'warning'
+  title?: string
+  borderColor?: string
```
mickaelzhang, line 37: `borderColor` shouldn't be customisable I think? It should be mapped with the type?

EDIT: It seems it's not used at all now, you can remove it I think?

src/SupportMessage/_tests_/SupportMessage.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+import 'jest-styled-components'
+
+import { SupportMessage, SupportMessageProps } from '../SupportMessage'
+
+const supportMessageProps: SupportMessageProps = {
+  description: 'Type info support message',
+  type: 'info',
+}
+
+describe('SupportMessage component', () => {
```
mickaelzhang, line 12: No need to have it inside the `_tests_` folder here, I don't think it brings much in this context, it made sense for snapshot as it create quite a few files

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,97 @@
...
+    margin-bottom: 24px;
+    padding: 16px;
+    display: flex;
+    font-family: ${theme.font.system};
+    color: ${theme.colors.secondary};
+    font-size: 12px;
+
+    span {
+      margin: 0 16px 0 0;
+    }
```
mickaelzhang, line 87: Can you wrap the `description` with a component instead?

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,97 @@
...
+  ({ type }) => css`
+    align-items: center;
+    background-color: ${styles[type].backgroundColor};
+    box-sizing: border-box;
+    ${type === 'info-outline' && `border: 1px solid ${theme.colors.secondary}`};
+    border-radius: 8px;
+    margin-bottom: 24px;
+    padding: 16px;
+    display: flex;
+    font-family: ${theme.font.system};
```
mickaelzhang, line 81: I think this should be the circular font? Not the system one

src/SupportMessage/SupportMessage.tsx
mickaelzhang: Not sure if you didn't push the change but it's still not 100%

src/SupportMessage/Collection.tsx
```diff
@@ -0,0 +1,88 @@
...
+      </ContainerBox>
+    </Box>
+  )
+}
+
+const ContainerBox = styled(Box)`
+  margin-bottom: 16px;
+  display: flex;
+  align-items: baseline;
+  max-width: 600px;
```
mickaelzhang, line 82: It would be better to remove the max-width or increase it, `300px` to show the SupportMessage is quite narrow

src/SupportMessage/Collection.tsx
```diff
@@ -0,0 +1,88 @@
...
+        />
+      </ContainerBox>
+    </Box>
+  )
+}
+
+const ContainerBox = styled(Box)`
+  margin-bottom: 16px;
+  display: flex;
+  align-items: baseline;
```
mickaelzhang, line 81: `center` is probably what you want but currently the margin that is inside `SupportMessage` messes with the spacing, when you've removed it `center` should work fine

```suggestion
  align-items: center;
```

mickaelzhang (commented):
I was checking the Figma and I think the icon should not be centered
- **Storybook**
Screenshot 2022-06-01 at 10 51 04

- **Figma**
Screenshot 2022-06-01 at 10 51 21

I've also added a new utility `useDeprecatedMessage`, can you maybe add it to the old `Message` component so that when one is on the page, we have a small message in the console?

src/SupportMessage/SupportMessage.tsx
githab666: Removed and wrapped content with styled component `ContentBox` instead

package.json
githab666: Already rebased?

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,104 @@
...
+const Wrapper = styled.div<IWrapper>(
+  ({ type }) => css`
+    align-items: center;
+    background-color: ${styles[type].backgroundColor};
+    ${type === 'info-outline' && `border: 1px solid ${theme.colors.secondary}`};
+    border-radius: 8px;
+    padding: 16px;
+    display: flex;
+    color: ${theme.colors.secondary};
+    font-size: 12px;
```
mickaelzhang, line 90: Can you move this and create a `Description` styled component with all the style link to the description?
Even though, it doesn't make a difference here, later on if we add some new styling, it will create less issue as all the styling are applied to only the part that we want

mickaelzhang, line 90: The font-size should be 14px I think from Figma?

mickaelzhang (changes requested):
Can you use the hook `useDeprecatedMessage` in the old `Message` component? It will help us identify when a legacy component is used

src/SupportMessage/SupportMessage.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Icon } from '../Icon'
+import { theme, Color } from '../theme'
+
+type StylesItem = {
+  iconColor: Color
+  backgroundColor: string
+  icon: string
+}
+
+const styles: Record<Type, StylesItem> = {
```
mickaelzhang, line 15: I think you meant `SupportMessageType` here?

src/Message/Message.tsx
```diff
@@ -4,6 +4,7 @@ import styled, { css } from 'styled-components'
 import { Box } from '../Box'
 import { Icon } from '../Icon'
 import { theme } from '../theme'
+import { useDeprecatedWarning } from '../utils/deprecated'
```
mickaelzhang, line 8: Could we have the `@deprecated` tag on the `Message` component as well so it appear as such in VS code?

src/SupportMessage/Collection.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <Box flex direction="column">
+      <ContainerBox>
+        <Label tag="span" typo="header-small">
+          Type info-outline
+        </Label>
+        <SupportMessage
+          type="info-outline"
+          description="Info support message"
+        />
+      </ContainerBox>
```
mickaelzhang, line 19: Is it normal that `info-outline` don't have an example with a title?

githab666: Yes doesn't have title example in designs + this is just displaying examples to user in Storybook

mickaelzhang (approved):
LGTM, great work 🙌

PR #1943, opened 2022-05-25
Typography story

fleonard:
## Screenshot / video

Screenshot 2022-05-25 at 09 42 48

## What does this do?

Add story for typography

### DIFF
--- a/src/Text/Text.stories.tsx
+++ b/src/Text/Text.stories.tsx
@@ -1,14 +1,74 @@
 import React from 'react'
+import styled from 'styled-components'
 import { Text, TextProps } from './Text'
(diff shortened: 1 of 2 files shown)


src/Text/Text.stories.tsx
```diff
@@ -30,3 +90,5 @@ WithTitle.args = {
   tag: 'p',
   title: 'a11y title tooltip',
 }
+
+export const TypographyCollection = TypographyTemplate.bind({})
```
mickaelzhang, line 94: No need for Typography part? We're already in the Text part, we can call it "Collection"

mickaelzhang (approved):

PR #1948, opened 2022-05-26
Make id optional and create id utility to generate them

mickaelzhang:
## What does this do?

Make `id` optional, in most cases, we don't really care about the `id` itself. We provide it because of accessibility purpose (to link the label to the input)
The only things that's important is that it's unique across a page


src/ActionDropdown/ActionDropdown.tsx
```diff
@@ -8,9 +8,7 @@ import { List, ActionListItem } from './List'
 import { Color, theme } from '../theme'
 
 export type ActionDropdownProps = {
-  /** ID, usually used for tests  */
-  id: string
-  /** className attribute to apply classes from props */
```
mickaelzhang, line 13: Just removing some comments that aren't necessary

src/utils/id.ts
```diff
@@ -0,0 +1,18 @@
+import { useMemo } from 'react'
+
+let idCount = 0
+
+export const smoresIdPrefix = 'MM_SMORES'
+
+const createId = (): string => {
+  const id = `${smoresIdPrefix}_${idCount}`
```
mickaelzhang, line 8: Added a prefix so we don't clash with any existing `id`

fleonard (approved):

PR #1978, opened 2022-06-09
stop always allowing today's date to be picked in datepicker

liampuk:
**Allowing 31 days to be selected from 15th June**

before:
Screenshot 2022-06-09 at 09 06 51

after:
Screenshot 2022-06-09 at 09 07 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,8 @@
-## [2.1.26] - 2022-06-09
-### Changed
-- Stop always allowing today's date to be picked
-
 ## [2.1.25] - 2022-05-27
 ### Added
 - Make `id` optional on form fields
 - Creates new Support Message component with types: 'info', 'info-outline', 'alert' and 'warning'
+- Stop always allowing today's date to be picked
 
 ## [2.1.24] - 2022-05-25
 ### Added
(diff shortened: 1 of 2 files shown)


fleonard:
Any extra context on this change? 
In most use cases we need to be able to select today as date (eg MTAs and new Inception Date in signup)

liampuk:
> Any extra context on this change? In most use cases we need to be able to select today as date (eg MTAs and new Inception Date in signup)

This is normally really weird behaviour as we generally want to allow selectable days after a certain date, but today's date is always visible. Is this something we'd want to have as an optional prop?

fleonard:
> > Any extra context on this change? In most use cases we need to be able to select today as date (eg MTAs and new Inception Date in signup)
> 
> This is normally really weird behaviour as we generally want to allow selectable days after a certain date, but today's date is always visible. Is this something we'd want to have as an optional prop?
> 
> @fleonard

@liampuk I guess it depends what we use as `fromDate`. We should hide today only if `fromDate` is not today otherwise should be included and active

liampuk:
@fleonard yes that is how it should work, I'll check fromDate is inclusive of today's date in those places

mickaelzhang (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [2.1.26] - 2022-06-09
+### Changed
+- Stop always allowing today's date to be picked
+
```
mickaelzhang, line 4: Need need to add a new version, you can add your change to `2.1.25`, it hasn't been released yet

PR #1979, opened 2022-06-09
Datepicker - Shift day by default

mickaelzhang:
## Screenshot / video

## What does this do?

- Give the reviewers context

### DIFF
--- a/src/Datepicker/DatesList.tsx
+++ b/src/Datepicker/DatesList.tsx
@@ -8,7 +8,6 @@ import { Day } from './types'
 
 type Props = {
   items: Day[]
-  firstDayShift: boolean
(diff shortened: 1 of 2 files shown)


src/Datepicker/DatesList.tsx
```diff
@@ -8,7 +8,6 @@ import { Day } from './types'
 
 type Props = {
   items: Day[]
-  firstDayShift: boolean
```
mickaelzhang, line 11: Are we good with removing the prop completely?
From the discussion that we had on Slack, having the calendar-like layout should be the default and I can't see a case where the other one is needed

fleonard (approved):

PR #2019, opened 2022-06-29
BgGradient pink tag

githab666:
## Screenshot
Screenshot 2022-06-29 at 17 46 21

## What does this do?
- Adds new boolean prop for styled gradient tag using `bgGradient`
- Adds to stories

### DIFF
--- a/src/Tag/Tag.tsx
+++ b/src/Tag/Tag.tsx
@@ -13,16 +13,16 @@ export type DefaultProps = {
 
 type GradientProps =
   | {
-      // if bgGradient === false
-      bgColor?: Color
-      borderColor?: Color
-      bgGradient?: false
-    }
-  | {
-      // if bgGradient === true
+      // no gradient
       bgColor: Color
       borderColor: Color
-      bgGradient: true
+      bgGradient?: boolean
+    }
+  | {
+      // with gradient
+      bgColor?: Color
+      borderColor?: Color
+      bgGradient: boolean
     }
 
 export type TagProps = DefaultProps & GradientProps
--- a/src/Tag/Tag.tsx
+++ b/src/Tag/Tag.tsx
@@ -8,9 +8,10 @@ export type TagProps = {
   label: string
   className?: string
   color: Color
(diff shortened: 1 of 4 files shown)


src/Tag/Tag.tsx
```diff
@@ -20,20 +21,30 @@ export const Tag: FC<TagProps> = ({
...
 const Wrapper = styled.div<WrapperProps>`
-  background-color: ${({ bgColor }) => theme.colors[bgColor]};
+  background-color: ${({ bgColor }) => bgColor && theme.colors[bgColor]};
+  background: ${({ bgGradient }) =>
+    bgGradient &&
+    `linear-gradient(90deg, rgba(247, 46, 73, 1) 0%, rgba(246, 148, 210, 1) 100%)`};
   border-radius: 8px;
-  border: 1px solid ${({ borderColor }) => theme.colors[borderColor]};
+  border: ${({ borderColor }) =>
+    borderColor && `1px solid ${theme.colors[borderColor]}`};
```
fleonard, line 47: if we remove the border for the gradient one I think we still need to take into consideration the size of the element? If we wanna keep always the same height I think we either need to tweak the height or have some sort of transparent border?

src/Tag/Tag.tsx
```diff
@@ -4,39 +4,63 @@ import styled from 'styled-components'
...
+      // if bgGradient === false
+      bgColor?: Color
+      borderColor?: Color
+      bgGradient?: false
+    }
+  | {
+      // if bgGradient === true
+      bgColor: Color
+      borderColor: Color
+      bgGradient: true
```
fleonard, line 25: I don't think this works. It should be
```
| {
      // no gradient
      bgColor: Color
      borderColor: Color
      bgGradient?: boolean
    }
  | {
      // with gradient
      bgColor?: Color
      borderColor?: Color
      bgGradient: boolean
}
```

We are trying to define which props are mandatory and which one are optional based on gradient vs solid colours

fleonard (approved):

PR #2023, opened 2022-06-30
Create field component

Jmuskett:
## Screenshot / video

using updated components, ready state

image

error state

image

NB alignment on the search button and address look up, this will require adding

```
display: flex,
align-items: center;
```
to `addressSearchWrapper` in AP - which I will do in a follow up PR

## What does this do?

- Creates a field component that takes an optional prop of "renderAsTitle"
- if renderAsTitle === true, then the label will be rendered in large/bold text, otherwise it will use the existing label style
- Field component is now used to wrap all input components in smores
- In addition, I've tidied up some of the code by removing unnecessary comments, and removing the use of `I` prefix on interfaces.


fleonard:
@Jmuskett is it just me or the required `*` in the `yes/no` form field is wrong? Shouldn't it be at the end of the copy rather than at the end of the container?

src/Dropdown/Dropdown.tsx
```diff
@@ -80,8 +54,8 @@ export type DropdownProps = DefaultProps & TruncateProps
 export const Dropdown = forwardRef(function Dropdown(
   {
     id: idProp,
-    className = '',
     label,
+    className = '',
```
josianed, line 58: is it necessary to default this to empty?

src/Field/Field.tsx
```diff
@@ -0,0 +1,141 @@
...
+  className?: string
+  id: string
+  error?: boolean // nb this needs to be optional for legacy support - but we will want to enforce in  the future
+  label?: string
+  outlined?: boolean
+  value: string
+  trailingIcon?: string
+  errorMsg?: string
+  dropdownKey?: string // nb only used by dropdown component but required on content in that context
+  required?: boolean // also optional for now to maintain backwards compaibility
```
josianed, line 23: what do you mean by 'for now'? is there plans to change the way this is implemented?

src/Field/Field.tsx
```diff
@@ -0,0 +1,141 @@
...
+  errorMsg,
+  dropdownKey,
+  required,
+  showCaret,
+  fullHeight = false,
+}: Props) => {
+  return (
+    <Container className={className}>
+      {label && (
+        <Box mb={outlined ? '4px' : '0px'}>
```
josianed, line 46: I think the spacing issue between the asterisk and the title on the `Confirmation` component needs to be addressed here, not sure where exactly without testing

src/SearchInput/SearchInput.tsx
```diff
@@ -77,13 +49,11 @@ export const SearchInput: FC<SearchInputProps> = ({
     const value = e.currentTarget.value
 
     if (value) {
-      // start filtering if the input has at least 2 characters
```
josianed, line 80: rip comments :(

src/SearchInput/SearchInput.tsx
```diff
@@ -77,13 +49,11 @@ export const SearchInput: FC<SearchInputProps> = ({
...
       if (value.length >= 2) {
         const filteredList = searchList.filter((el) =>
           el.label.toLowerCase().includes(value.toLocaleLowerCase()),
         )
 
-        // update the local state with the filtered results
```
josianed, line 86: you'll be missed :(

src/Field/Field.tsx
Jmuskett: good spot, my bad - I'll remove the comment

src/Field/Field.tsx
Jmuskett: @josianed happy for you to pick this up as referenced above, if you've have more context 👍🏻

src/Dropdown/Dropdown.tsx
Jmuskett: I'm not 100% sure, but I didn't want to risk something silently failing somewhere, that we only then find out about when a user tells us - so I kept the existing pattern in this instance.

src/Field/Field.tsx
```diff
@@ -0,0 +1,141 @@
...
+  error?: boolean // nb this needs to be optional for legacy support - but we will want to enforce in  the future
+  label?: string
+  outlined?: boolean
+  value: string
+  trailingIcon?: string
+  errorMsg?: string
+  dropdownKey?: string // nb only used by dropdown component but required on content in that context
+  required?: boolean
+  fullHeight?: boolean // used to make text areas full size
+}
```
mickaelzhang, line 25: A big part of those props are shared between the inputs, we could create a common interface that can be used across them all? I'm thinking of `label`, `id`, `error`, `errorMsg`, `required`, `renderAsTitle`


```ts
type CommonFieldProps = {
  id?: string
  label?: string
  ...
}

type TextInputProps = { ... } & CommonFieldProps 
```

src/Field/Field.tsx
```diff
@@ -0,0 +1,141 @@
...
+          ) : (
+            <Text tag="label" color="subtext" typo="label" htmlFor={id}>
+              {label}
+            </Text>
+          )}
+          {required && (
+            <Text tag="label" color="error" typo="label">
+              *
+            </Text>
+          )}
```
fleonard, line 58: wondering if we should move this inside the above Text and have it a span?
like
```
<Text tag="label" color="subtext" typo="label" htmlFor={id}>
{label}
{required && (
<Text tag="span" color="error" typo="label">
 *
</Text>
)}
</Text>
```

cause like this we are rendering two labels that's not ideal

src/TextInput/TextInput.tsx
```diff
@@ -71,28 +57,27 @@ export const TextInput = forwardRef(function TextInput(
     onInputChange,
     disabled = false,
     required = false,
+    renderAsTitle = false,
   }: TextInputProps,
```
mickaelzhang, line 61: We shouldn't have to declare props that are just passed to `Field` manually
I think we can just spread the rest
```suggestion
    disabled = false,
    ...fieldProps
  }: TextInputProps,
```

That way, with new `Field` API update, we wouldn't need to update any other components as everything should be handled by default

src/Textarea/Textarea.tsx
```diff
@@ -70,59 +54,54 @@ export const Textarea = forwardRef(function Textarea(
...
+        value={value}
+        onChange={(e: FormEvent<HTMLTextAreaElement>) => {
+          onChange && onChange(e.currentTarget.value)
+          onInputChange && onInputChange(e)
+        }}
+        maxLength={maxLength}
+        ref={ref}
+        onBlur={(e) => {
+          onBlur && onBlur(e)
+        }}
```
mickaelzhang, line 90:
```suggestion
        onBlur={onBlur}
```

src/Textarea/Textarea.tsx
```diff
@@ -70,59 +54,54 @@ export const Textarea = forwardRef(function Textarea(
...
+        error={error}
+        id={id}
+        name={name}
+        disabled={disabled}
+        resize={resize}
+        placeholder={placeholder}
+        value={value}
+        onChange={(e: FormEvent<HTMLTextAreaElement>) => {
+          onChange && onChange(e.currentTarget.value)
+          onInputChange && onInputChange(e)
```
mickaelzhang, line 84:
```suggestion
          onChange?.(e.currentTarget.value)
          onInputChange?.(e)
```

src/Textarea/Textarea.tsx
```diff
@@ -70,59 +54,54 @@ export const Textarea = forwardRef(function Textarea(
...
+    >
+      <StyledTextArea
+        error={error}
+        id={id}
+        name={name}
+        disabled={disabled}
+        resize={resize}
+        placeholder={placeholder}
+        value={value}
+        onChange={(e: FormEvent<HTMLTextAreaElement>) => {
```
mickaelzhang, line 82: This don't need to be typed as it is inferred by default
```suggestion
        onChange={(e) => {
```

src/TextInput/TextInput.tsx
```diff
@@ -1,45 +1,31 @@
...
+import { Field } from '../Field'
 import { useUniqueId } from '../utils/id'
 
-type DefaultProps = {
-  id?: string
+type Props = {
+  id: string
   className?: string
-  /** Input type for proper browser support */
+  ref?: RefObject<HTMLInputElement>
```
mickaelzhang, line 11: Adding `ref` here isn't needed, it's already declared using `forwardRef`

src/SearchInput/SearchInput.tsx
```diff
@@ -186,7 +160,16 @@ const InputBox = styled.div<IShowIcon>`
...
 `
 
-const Input = styled.input<ISearchInput>`
+interface Input extends UsesOutline {
+  id: string
+  name: string
+  value: string
+  selected: boolean
+  onKeyUp: (e: React.FormEvent<HTMLInputElement>) => void
+  onChange: (e: React.FormEvent<HTMLInputElement>) => void
```
mickaelzhang, line 169: Shouldn't those already be on `input` API? I don't think we need to declare them explicitly?

src/Field/Field.tsx
Jmuskett: It still wants to be in a `Text` tag so that we can apply the right colour, but I've changed the `typo` to base so that we're not creating multiple labels, in [e377d0](https://github.com/marshmallow-insurance/smores-react/pull/2023/commits/e377d04ddbca32b14a683c40ee60eb58a5e54dcf)

src/Field/Field.tsx
```diff
@@ -0,0 +1,141 @@
...
+  flex-direction: column;
+  position: relative;
+`
+
+const Content = styled.div<{
+  error?: boolean
+  outlined: boolean
+  value?: string
+  fullHeight?: boolean
+}>`
```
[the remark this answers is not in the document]
Jmuskett: Hmm, not sure I agree - since the idea behind this is to make sure we have uniform styling across all fields, I think it makes sense to enforce things like height and border on the field level. We could abstract into it's own component as you suggested, but I think at this point that fits into the over abstraction category.  It would also break the philosophy behind doing 

```
<Field>
<whateverInput/>
</Field>
```

as We'd now need a separate Content (or whatever it was called) wrapper as well

src/Field/Field.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { Box } from '../Box'
+import { theme } from '../theme'
+
+type Props = {
+  children: React.ReactElement
+  showCaret?: boolean
+  renderAsTitle?: boolean
+  className?: string
+  id: string
+  error?: boolean // nb this needs to be optional for legacy support - but we will want to enforce in  the future
```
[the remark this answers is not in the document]
Jmuskett: I think in the future we'll want to enforce passing of error states, no? Every field should have a disabled and error state IMO

src/Field/Field.tsx
Jmuskett: I'm not sure I follow - what's the benefit of doing that, since all the types are encapsulated in the Field component?

src/TextInput/TextInput.tsx
Jmuskett: I'm not sure I follow - with the existing pattern we don't need to update any existing components, as the  new prop is optional and false as default?

src/Textarea/Textarea.tsx
Jmuskett: I don't think that's right - TS linter complains that e is implict any, without the typing.

fleonard: yeah I think that's why we had the type in the first place

src/Field/Field.tsx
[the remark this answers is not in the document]
fleonard: There might be instances were we don't want to use the error of the form field but some other visuals so probably worth keeping it optional?

src/Field/Field.tsx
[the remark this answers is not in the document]
fleonard: Yeah I don't think they are input related. They are spacing and layout changes based on how the input behave so agree that should live in `Field`

src/Field/Field.tsx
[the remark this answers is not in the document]
Jmuskett: Fair, removed the comment in  [daede0e](https://github.com/marshmallow-insurance/smores-react/pull/2023/commits/daede0e441d68bf9cab27de06e62782e7491e161)

fleonard (dismissed):
LGTM 🙌🏻

josianed (approved):
ship it!

src/TextInput/TextInput.tsx
mickaelzhang: I think `ref` hasn't been removed?

src/Field/Field.tsx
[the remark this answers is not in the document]
mickaelzhang: I guess if it used only for input, we can keep it there but I feel that `Field` could be used for other types of control like the date picker or a custom components

src/Textarea/Textarea.tsx
Jmuskett: not on my machine 😅
image

mickaelzhang: Weird 🤔 The styled component itself should be rightfully typed so we shouldn't have any issue, what's the error message?

src/TextInput/TextInput.tsx
mickaelzhang: Yes you can just spread the props that aren't actually needed inside the input components:
```ts
const Component = ({ error, ...fieldProps }) => {
  return (
    <Field {...fieldProps}>
      <Input  error={error} />
    </Field>
  )
}
```

In future updates of `Field`, if we add a new prop, it's going to be passed automatically to all component that consumes `Field` and the typing would be correct as well since the types are shared

Jmuskett: Fixed in [8b4f6ed](https://github.com/marshmallow-insurance/smores-react/pull/2023/commits/8b4f6ed40ca0c108946ed7259b04ad958ba7315a)

fleonard (approved):

PR #2031, opened 2022-07-04
add fieldset component

Jmuskett:
## What does this do?

create fieldset component (not currently in use) for groups of form fields.


fleonard:
Should this be exported from the main index file? Otherwise we can't use it

PR #2036, opened 2022-07-05
Update SupportMessage description to accept React element

ellajay:
## What does this do?

Updates `SupportMessage` component so that the description can accept React elements


mickaelzhang (commented):
Could you add an example of usage in Storybook?

src/SupportMessage/SupportMessage.stories.tsx
```diff
@@ -21,4 +22,12 @@ const supportMessageArgs: SupportMessageProps = {
 
 Default.args = supportMessageArgs
 
+export const WithComponent = Template.bind({})
+
+WithComponent.args = {
+  type: 'info',
+  title: 'A SupportMessage using the Text component',
+  description: <Text tag="p">Some example text</Text>,
+}
```
mickaelzhang, line 31: Could you perhaps use your real use cases as an example on Storybook? I think it would make it clearer in which kind of context it is ok to use a custom element as a description because in this case it doesn't bring any value

src/SupportMessage/SupportMessage.stories.tsx
```diff
@@ -21,4 +22,12 @@ const supportMessageArgs: SupportMessageProps = {
 
 Default.args = supportMessageArgs
 
+export const WithComponent = Template.bind({})
```
mickaelzhang, line 25: `WithComponent` might not be descriptive enough, maybe `WithCustomDescription`?

src/SupportMessage/SupportMessage.stories.tsx
```diff
@@ -21,4 +23,17 @@ const supportMessageArgs: SupportMessageProps = {
...
+export const WithCustomDescription = Template.bind({})
+
+WithCustomDescription.args = {
+  type: 'info',
+  title: 'A SupportMessage using the Text component',
+  description: (
+    <Text tag="p">
+      Some text rendered using the Text component and a{' '}
+      <Link href={''}>Link</Link>
+    </Text>
```
mickaelzhang, line 35: The story doesn't look right, I think the `Text` here is messing with the styling
```suggestion
    <>
      Some text rendered using the Text component and a{' '}
      <Link href={''}>Link</Link>
    </>
```

The following should fix the style

mickaelzhang (approved):

PR #2055, opened 2022-07-11
Adds width for search container

githab666:
## Screenshot
Screenshot from using yalc on sign up across bugs:
Screenshot 2022-07-11 at 15 47 15

## What does this do?
- Adds width 100% to fix bug across sign up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 
+## [2.3.5] - 2022-07-11
+### Changed 
+- Width css for Field component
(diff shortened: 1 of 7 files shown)


fleonard:
@githab666 why the component has been reverted back and not using the Field component?

githab666:
@fleonard I only made change to the Field component and updated snapshots, just looking into it now as something weird happening with my PR

fleonard:
@githab666 yeah though was something odd going on!

PR #2060, opened 2022-07-12
Updates width of field

githab666:
## Screenshot 
Fix for search inputs fixing across sign up. When running yalc this example is fixed:
Screenshot 2022-07-11 at 15 47 15

## What does this do?
- Adds `width: 100%` to Field component
- Updates `CHANGELOG` and snapshots

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,12 +56,13 @@
       }
     },
     "node_modules/@ampproject/remapping": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.1.1.tgz",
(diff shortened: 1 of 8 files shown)


fleonard (dismissed):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mrshmllw/smores-react",
-  "version": "2.3.4",
+  "version": "2.3.5",
```
fleonard, line 3: if you bump in the same PR the release action will probably break

githab666: Oh i didn't know that... Will remove

package-lock.json
```diff
@@ -7436,23 +7436,6 @@
         "node": ">=10"
```
fleonard, line 7436: if you haven't bumped this shouldn't be in this PR

fleonard (approved):

PR #2074, opened 2022-07-15
Support message style

mickaelzhang:
## Screenshot / video

- **BEFORE**
Screenshot 2022-07-15 at 11 57 39

- **AFTER**
Screenshot 2022-07-15 at 12 00 35

## What does this do?

When using custom description, text were not handle properly

### DIFF
--- a/src/SupportMessage/Collection.tsx
+++ b/src/SupportMessage/Collection.tsx
@@ -3,86 +3,75 @@ import styled from 'styled-components'
 
 import { Text } from '../Text'
 import { Box } from '../Box'
-import { SupportMessage } from './SupportMessage'
+import { SupportMessage, SupportMessageProps } from './SupportMessage'
+
+const variationList: SupportMessageProps[] = [
+  {
+    type: 'info',
+    title: 'Additional information',
+    description: 'Information needs to be communicated to the user.',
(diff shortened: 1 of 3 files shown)


src/SupportMessage/Collection.tsx
```diff
@@ -3,86 +3,75 @@ import styled from 'styled-components'
...
-        <Label tag="span" typo="header-small">
-          Type warning - With title
-        </Label>
-        <SupportMessage
-          type="warning"
-          title="General warning title"
-          description="Warning support message"
-        />
-      </ContainerBox>
+      {variationList.map(({ type, title, description }) => {
```
mickaelzhang, line 34: Screenshot 2022-07-15 at 12 19 52

I've updated the view a bit and added some more concrete example

fleonard:
@ellajay I know you had to add the `link` for some work on LTV. Can you double check that it's fine to have it `secondary` and not `primary` in the design?

fleonard (changes requested):
you might need to rebase and update the Changelog since the version has been bumped

mickaelzhang:
@fleonard Should be good now

fleonard (approved):

PR #2084, opened 2022-07-19
Field component updates

josianed:
## Screenshot / video

Screenshot 2022-07-19 at 13 00 52

## What does this do?

- Update Field component title with correct tag and styles
- Add storybook example for rendering label as title
- Add htmlFor prop linked to input field

### DIFF
--- a/src/Field/Field.tsx
+++ b/src/Field/Field.tsx
@@ -39,7 +39,7 @@ export const Field = ({
       {label && (
         <Box mb={{ custom: outlined ? 4 : 0 }}>
           {renderAsTitle ? (
-            <Title>{label}</Title>
+            <Title htmlFor={id}>{label}</Title>
           ) : (
             <Text tag="label" color="subtext" typo="label" htmlFor={id}>
               {label}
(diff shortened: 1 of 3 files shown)


src/Field/Field.tsx
```diff
@@ -117,7 +117,9 @@ const ErrorBox = styled.span`
   font-size: 12px;
 `
 
-const Title = styled.h3`
+const Title = styled.label`
+  font-size: 16px;
+  line-height: 26px;
```
fleonard, line 122: are these coming from a specific design? @Jmuskett do you mind double check that these values are good with Direct Flow?

fleonard: Looking at Figma designs for direct flow it should follow this style
Screenshot 2022-07-19 at 14 58 59

Jmuskett: I think we want it uniform across everything so it makes sense to use figma as the source of truth I think

josianed: thanks! updated to reflect this.

fleonard (approved):

fleonard (approved):

PR #2086, opened 2022-07-19
Fix search input

fleonard:
## Screenshot / video
Before
Screenshot 2022-07-19 at 15 41 15
After
Screenshot 2022-07-19 at 15 34 11

## What does this do?
`flex` in the field container was breaking the size of the search input

### DIFF
--- a/src/SearchInput/SearchInput.tsx
+++ b/src/SearchInput/SearchInput.tsx
@@ -192,7 +192,7 @@ const StyledResultsContainer = styled.div<ResultsContainer>`
   ${({ absolutePosition }) => absolutePosition && 'position: absolute;'}
   width: 100%;
   visibility: ${({ show }) => (show ? 'visible' : 'hidden')};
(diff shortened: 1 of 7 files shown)


src/SearchInput/SearchInput.tsx
```diff
@@ -192,7 +192,7 @@ const StyledResultsContainer = styled.div<ResultsContainer>`
   ${({ absolutePosition }) => absolutePosition && 'position: absolute;'}
   width: 100%;
   visibility: ${({ show }) => (show ? 'visible' : 'hidden')};
-  ${({ outlined }) => outlined && 'left: 0px; top: 93%;'};
+  ${({ outlined }) => outlined && 'left: 0px; top: 90%;'};
```
fleonard, line 195: better hiding the bottom border corners

mickaelzhang: Shouldn't we remove the border-radius at the bottom when the popover is open instead of doing this?

mickaelzhang (approved):

PR #2092, opened 2022-07-20
Migrate NumberInput to use Field component

fleonard:
## What does this do?

If you don't mind checking out the branch and test it might be a good thing cause it has **A LOT** of edge cases 😢

### DIFF
--- a/src/NumberInput/NumberInput.tsx
+++ b/src/NumberInput/NumberInput.tsx
@@ -1,81 +1,71 @@
-import React, { FormEvent, FC, MouseEvent } from 'react'
+import React, { FormEvent, forwardRef, ForwardedRef, MouseEvent } from 'react'
 import styled from 'styled-components'
 
 import { Box } from '../Box'
 import { Text } from '../Text'
 import { Icon } from '../Icon'
+import { Field, CommonFieldTypes } from '../Field'
 
 import { theme } from '../theme'
 import { useUniqueId } from '../utils/id'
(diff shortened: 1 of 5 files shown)


src/TextInput/TextInput.tsx
```diff
@@ -60,7 +61,7 @@ export const TextInput = forwardRef(function TextInput(
       outlined={outlined}
       value={value}
     >
-      <>
+      <Box flex>
```
fleonard, line 64: this is needed to fix the trailing icon that's now broken because we remove `flex` from the field content

src/NumberInput/NumberInput.tsx
```diff
@@ -220,109 +214,115 @@ export const NumberInput: FC<NumberInputProps> = ({
...
   border-color: ${({ error }) =>
     theme.colors[`${error ? 'error' : 'outline'}`]};
-  display: flex;
-  align-items: center;
-  height: 32px;
+  color: ${({ error }) => theme.colors[`${error ? 'error' : 'secondary'}`]};
+  font-size: 16px;
+  width: 100%;
+  outline: none;
   cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
```
mickaelzhang, line 241: Not part of your PR but `pointer` is a bit weird for an input? Shouldn't we use `text` instead?

mickaelzhang (approved):

PR #2105, opened 2022-07-27
Fix outline color on accordion component 

josianed:
- Border color was wrongly specified during the color migration
- Updated full border to match bottom border only


src/Accordion/Accordion.tsx
```diff
@@ -70,7 +70,7 @@ const Wrapper = styled(Box)<Omit<IAccordion, 'isOpen'>>(
```
fleonard, line 70: do you also wanna update the changelog if you are going to release this now?

fleonard (approved):

PR #2111, opened 2022-08-01
Remove unused codacy badge

lapostoj:
Mentioning it in the call earlier made me realise we still have some mentions in places which are way outdated.

Travis CI must also not be relevant anymore.
And I'm not sure about that David thing, but it seems we're not getting data anymore?


fleonard (approved):
yeah I think we can remove that david thing since I have no idea what it is

lapostoj:
@fleonard yes build was still checking travisCI it seems, the one I'm removing as well. Not sure if there is a github action one and it it's worth adding or not

fleonard:
@lapostoj yeah I was actually thinking about looking if we can add something about GitHub Actions 👍🏻

fleonard (approved):

PR #2116, opened 2022-08-02
New typography

fleonard:
## Screenshot / video

[RFC](https://www.notion.so/getmarshmallow/Typography-51b84dc66dac4ef3a82f4b3b6cac97e5)

## What does this do?

- Update `Text` component to support new typography
- Deprecate old styles without breaking changes
- Add new font


src/Text/Text.tsx
```diff
@@ -18,11 +18,45 @@ interface IText {
   cursor: string
 }
 
+/**
+ * @deprecated Use only new Typo
+ */
+
+export type DeprecatedTypo =
```
fleonard, line 25: this is not used but I kept it here just so we don't forget the old names

src/Text/Text.stories.tsx
```diff
@@ -2,32 +2,33 @@ import React from 'react'
...
 import { Box } from '../Box'
+import { Typo } from './Text'
 
 export default {
   title: 'Text',
   component: Text,
 }
 
 const typos = [
-  'header-large',
```
fleonard, line 12: wondering if we should keep displaying the deprecated typos in a separate story?

src/Field/Field.tsx
```diff
@@ -39,11 +39,17 @@ export const Field = ({
       {label && (
         <Box mb={{ custom: outlined ? 4 : 0 }}>
           {renderAsTitle ? (
-            <Title htmlFor={id}>{label}</Title>
+            <Text tag="label" typo="heading-small" htmlFor={id} mb="16px">
+              {label}
+            </Text>
           ) : (
-            <Text tag="label" color="subtext" typo="label" htmlFor={id}>
+            <Text tag="label" typo="label" color="subtext" htmlFor={id}>
```
mickaelzhang, line 46: We don't need any margin bottom for this case?

src/Field/Field.tsx
```diff
@@ -39,11 +39,17 @@ export const Field = ({
...
               {label}
-              {required && <Asterisk>*</Asterisk>}
+              {required && (
+                <Text tag="span" typo="body-small" color="error">
+                  *
+                </Text>
+              )}
```
mickaelzhang, line 52: This isn't needed for the title version?

src/Link/Link.tsx
```diff
@@ -6,13 +6,16 @@ import { focusOutline } from '../utils/focusOutline'
...
+type LinkTypo = 'regular' | 'small'
+
 export type LinkProps = {
   className?: string
   href: string
   onClick?: MouseEventHandler
   openInNewTab?: boolean
   children?: ReactNode
   download?: boolean
+  typo?: LinkTypo
```
mickaelzhang, line 18: I don't think we want to make it customisable, it should inherit the styling from its parent instead?

src/Tag/Tag.tsx
```diff
@@ -2,15 +2,15 @@ import React, { FC } from 'react'
...
+import { Text, DeprecatedTypo, Typo } from '../Text'
 import { Color, theme } from '../theme'
 import { Box } from '../Box'
 
 export type DefaultProps = {
   label: string
   className?: string
   color: Color
-  typo?: string
+  typo?: Typo | DeprecatedTypo
```
mickaelzhang, line 13: I wasn't even aware that we could change the `typo` here
Since this component is bound to be deprecated, should we leave it at `string`?
I'm pretty sure that we don't want to be able to update the typo for the next version

src/Text/Text.tsx
```diff
@@ -222,8 +334,6 @@ const Container = styled(Box)<IText>(
       letter-spacing: 0.75px;
     `}
 
-    margin: 0;
-    padding: 0;
```
mickaelzhang, line 226: Isn't this gonna create some issue on some repos? Did you check repos reset files to be sure that everything is properly setup?

src/Field/Field.tsx
fleonard: I've just replaced old typos with the new one design hasn't changed so if it was fine before should be fine now

src/Field/Field.tsx
fleonard: I haven't touched anything is this component. It was like this even before and I don't know how it works from a design point of view since I never saw an example with title with required

src/Link/Link.tsx
fleonard: typography has 2 new styles for the link and this will never be wrapped by a `Text` component so we need them here

src/Tag/Tag.tsx
fleonard: yeah fair enought

src/Text/Text.stories.tsx
fleonard: I'll add a new collection

src/Text/Text.tsx
fleonard: yeah all repos are resetting margin and padding in global style and these two were breaking the new margin props

src/Text/Text.tsx
```diff
@@ -66,9 +100,111 @@ Text.displayName = 'Text'
 
 const Container = styled(Box)<IText>(
   ({ align, color, cursor, typo }) => css`
-    /** PREDEFINED TYPOGRAPHY STYLES */
+    /** TYPOGRAPHY STYLES */
+
+    ${typo === 'hero-alternate' &&
+    css`
+      font-size: 40px;
```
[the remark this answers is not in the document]
fleonard: I don't think it's worth it. these will never change and I've ported them over from Figma double checking with Alistar that is all matching

src/Text/Text.tsx
```diff
@@ -66,9 +100,111 @@ Text.displayName = 'Text'
...
+      font-family: 'ClearfaceITC';
+      font-weight: ${theme.font.weight.bold};
+      line-height: 44px;
+
+      @media (min-width: 768px) {
+        font-size: 56px;
+        line-height: 64px;
+      }
```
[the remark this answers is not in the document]
fleonard: some of the styles has mobile/desktop - it's all set up correctly in Figma from design so I think they are aware.
Also discussing with Alistar and Georgia we should never override any kind of typo. We should always use one of these styles

src/Text/Text.tsx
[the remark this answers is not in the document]
mickaelzhang: Not sure what you mean by more than a string, wouldn't this work?

```ts
const fontStyleMapping: Record<Typo, string> = {
  hero: `
      font-size: 32px;
      font-weight: ${theme.font.weight.medium};
      line-height: 40px;
      @media (min-width: 768px) {
        font-size: 40px;
        line-height: 44px;
      }
  `,
   ...
}
```

By having a mapping, we wouldn't need to maintain a separate array for the documentation as well since we can just infer it from the mapping:
```ts
const typos = Object.keys(fontStyleMapping)
```
Storybook would always be in sync with the codebase this way

src/Link/Link.tsx
mickaelzhang: Looking at Figma, it seems to match with Text typo `body-regular` and `body-small`

I don't see any reason to be able to customise the `Link` size, except inside a paragraph so I'm guessing that the size `small` for `Link` should only be used when inside `body-small` and the default and the one that should be used in most situation is `regular`

Could we confirm with design?

fleonard: ok, let me check with Alistair. I still think that we need those sizes here since if we use the link outside of a paragraph we don't have any option to set the font and I don't think it should be wrapped by a `Text` component just to get the typo

src/Text/Text.tsx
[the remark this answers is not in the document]
fleonard: I don't think I get how that can be better than the way it's currently set up. Storybook is just rendering texts based on the `typo` how can it go out of sync currently? I think it's much easier to read if we have the actual styles in the css rather than in a mapping file

src/Link/Link.tsx
mickaelzhang: Yeah definitely, my comment was with the assumption that we never need Link `small`, except in `body-small`
And Link `regular` should be used in most situation

If there is a need for Link `small` tho, I'm good with adding a prop for it

src/Text/Text.tsx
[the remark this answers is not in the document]
fleonard: oh ok now I got what you ment yeah I suppose we can do it that way

ellajay (approved):

src/Text/Text.tsx
```diff
@@ -64,11 +99,31 @@ export const Text: FC<TextProps> = forwardRef<HTMLElement, TextProps>(
...
+    ${styles}
+  `
+}
+
 const Container = styled(Box)<IText>(
   ({ align, color, cursor, typo }) => css`
-    /** PREDEFINED TYPOGRAPHY STYLES */
+    /** TYPOGRAPHY STYLES */
+
+    ${isNewTypo(typo) && generateTypo(typo)}
```
mickaelzhang, line 124: You should be able to inline `fontStyleMapping` here
```suggestion
    ${isNewTypo(typo) && fontStyleMapping[font]}
```

src/Link/Link.tsx
fleonard: @mickaelzhang I've checked with Alistair and we agreed that might be safer to keep the style as a prop cause for now there's not a case where we use links outside paragraphs but we can't 100% be sure that in the future there won't be.
So I'd rather keep the prop now and have to always add it if we need a small Link then to have to go back and change it across all repos in future

mickaelzhang: If we don't have a case for having it outside of paragraph for now, wouldn't it be safer to not have a prop?
Adding a prop is easy compared to remove it and check if there hasn't been bad usage across app?

fleonard: It's currently defaulting to `regular` anyway so we'll only need to add the `small` in case we need it and I'd rather have the prop from the start in case we start using standalone links outside paragraphs in the future rather than having to wrap them in a `Text` to inherit the style

mickaelzhang (approved):

PR #2118, opened 2022-08-03
add assistive text to Field

Jmuskett:
image

- adds a new prop to Field component, assistiveText
- allows subtext/body copy to be added below a question label for more detail

### DIFF
--- a/src/Field/Field.tsx
+++ b/src/Field/Field.tsx
@@ -56,13 +56,15 @@ export const Field = ({
           )}
         </Box>
       )}
-      <Box>
-        {assistiveText && (
-          <Text tag="p" typo="body" color="subtext">
+
+      {assistiveText && (
+        <Box>
+          <Text tag="p" color="subtext">
             {assistiveText}
           </Text>
-        )}
-      </Box>
+        </Box>
+      )}
+
       <Content
         fullHeight={fullHeight}
         value={value}
--- a/src/Field/Field.tsx
+++ b/src/Field/Field.tsx
@@ -46,7 +46,7 @@ export const Field = ({
                 tag="label"
                 typo="heading-small"
                 htmlFor={id}
-                mb={{ custom: 4 }}
+                mb={assistiveText ? { custom: 4 } : undefined}
               >
                 {label}
               </Text>
--- a/src/Field/Field.tsx
+++ b/src/Field/Field.tsx
@@ -39,10 +39,15 @@ export const Field = ({
   return (
     <Container className={className} {...marginProps}>
       {label && (
-        <Box mb={{ custom: outlined ? 4 : 0 }}>
+        <>
           {renderAsTitle ? (
             <Box mb="16px">
-              <Text tag="label" typo="heading-small" htmlFor={id}>
+              <Text
+                tag="label"
+                typo="heading-small"
+                htmlFor={id}
+                mb={{ custom: 4 }}
+              >
                 {label}
               </Text>
 
@@ -53,16 +58,18 @@ export const Field = ({
               )}
             </Box>
           ) : (
-            <Text tag="label" typo="label" color="subtext" htmlFor={id}>
-              {label}
-              {required && (
-                <Text tag="span" typo="body-small" color="error">
-                  *
-                </Text>
-              )}
-            </Text>
+            <Box mb={{ custom: outlined ? 4 : 0 }}>
+              <Text tag="label" typo="label" color="subtext" htmlFor={id}>
+                {label}
+                {required && (
+                  <Text tag="span" typo="body-small" color="error">
+                    *
+                  </Text>
+                )}
+              </Text>
+            </Box>
           )}
-        </Box>
+        </>
       )}
 
       <Content
--- a/src/Field/Field.tsx
+++ b/src/Field/Field.tsx
@@ -41,9 +41,17 @@ export const Field = ({
       {label && (
         <Box mb={{ custom: outlined ? 4 : 0 }}>
           {renderAsTitle ? (
-            <Text tag="label" typo="heading-small" htmlFor={id} mb="16px">
-              {label}
-            </Text>
+            <Box mb="16px">
+              <Text tag="label" typo="heading-small" htmlFor={id}>
+                {label}
+              </Text>
+
+              {assistiveText && (
+                <Text tag="p" color="subtext">
+                  {assistiveText}
+                </Text>
+              )}
+            </Box>
           ) : (
             <Text tag="label" typo="label" color="subtext" htmlFor={id}>
               {label}
@@ -55,12 +63,6 @@ export const Field = ({
             </Text>
           )}
         </Box>
-      )}
-
-      {assistiveText && (
-        <Text tag="p" color="subtext">
-          {assistiveText}
-        </Text>
       )}
 
       <Content
--- a/src/Field/Field.tsx
+++ b/src/Field/Field.tsx
@@ -15,6 +15,7 @@ interface FieldProps extends CommonFieldTypes {
   trailingIcon?: string
   dropdownKey?: string // nb only used by dropdown component but required on content in that context
   fullHeight?: boolean // used to make text areas full size
+  assistiveText?: string
 }
 
 export const Field = ({
@@ -32,6 +33,7 @@ export const Field = ({
   required,
   showCaret,
   fullHeight = false,
+  assistiveText,
   ...marginProps
 }: FieldProps) => {
   return (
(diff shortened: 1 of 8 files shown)


src/Field/Field.tsx
```diff
@@ -54,6 +56,13 @@ export const Field = ({
           )}
         </Box>
       )}
+      <Box>
+        {assistiveText && (
+          <Text tag="p" typo="body" color="subtext">
```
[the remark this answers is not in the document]
fleonard: Also body doesn't exist as part of the new typo. It's eaither `body-regular` and in that case you don't need the `typo` since it's default or `body-small`

fleonard (commented):
Are we not adding it to the `Confirmation` component? It currently doesn't use `Field` but we might need to have a bigger conversation about it

src/Field/types/commonFieldTypes.ts
```diff
@@ -9,4 +9,5 @@ export interface CommonFieldTypes extends MarginProps {
   renderAsTitle?: boolean
   className?: string
   outlined?: boolean
+  assistiveText?: string
```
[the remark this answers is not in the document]
Jmuskett: yes

src/Field/Field.tsx
```diff
@@ -56,13 +56,15 @@ export const Field = ({
...
-          <Text tag="p" typo="body" color="subtext">
+
+      {assistiveText && (
+        <Box>
+          <Text tag="p" color="subtext">
             {assistiveText}
           </Text>
-        )}
-      </Box>
+        </Box>
```
fleonard, line 65: do we actually need the Box at all? Have you also checked that margins are fine? I also believe that this case will only apply if we have `renderAsTitle`? wondering if we should combine them cause I don't think we'll ever have a case were we have `assistiveText` and no `renderAsTitle`

fleonard (commented):
We also need to add it to the `FieldGroup` right?

src/Field/Field.tsx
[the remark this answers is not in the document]
mickaelzhang: @Jmuskett Not sure if you saw this comment, did you check margins between the title and assistive text? And also how margin between label/assistive text and input under it?

Jmuskett:
> We also need to add it to the `FieldGroup` right?

We don't, the FieldSet is just a wrapper for adding 

`fieldset` and `legend` to a group of fields, so the individual fields will still have access to the new prop, e.g

```
<FieldSet legend ='answer some questions>
<Textinput assistiveText="foo />
</FieldSet>
```

fleonard (dismissed):

src/Field/Field.tsx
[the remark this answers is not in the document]
Jmuskett: Added in [616b2a](https://github.com/marshmallow-insurance/smores-react/pull/2118/commits/616b2a5e21517f733b6649623a18b1642444226d)

src/Field/Field.tsx
Jmuskett: Moved it into the `renderAsTitle` conditional in [616b2a](https://github.com/marshmallow-insurance/smores-react/pull/2118/commits/616b2a5e21517f733b6649623a18b1642444226d)

src/TextInput/TextInput.stories.tsx
```diff
@@ -140,6 +140,21 @@ Required.args = {
...
+AssistiveText.args = {
+  id: 'textInput',
+  name: 'textInput',
+  placeholder: 'Enter name',
+  onChange: noop,
+  onInputChange: noop,
+  onBlur: noop,
+  label: 'Please enter your name',
+  outlined: true,
+  required: true,
```
mickaelzhang, line 154: You forgot to add `renderAsTitle` for the story
```suggestion
  required: true,
  renderAsTitle: true,
```

src/Field/Field.tsx
```diff
@@ -32,16 +33,25 @@ export const Field = ({
...
-            <Text tag="label" typo="heading-small" htmlFor={id} mb="16px">
-              {label}
-            </Text>
+            <Box mb="16px">
+              <Text tag="label" typo="heading-small" htmlFor={id}>
+                {label}
+              </Text>
+
+              {assistiveText && (
+                <Text tag="p" color="subtext">
```
mickaelzhang, line 50: We should have 4px between title and assistive text

src/Field/Field.tsx
```diff
@@ -32,16 +33,25 @@ export const Field = ({
   required,
   showCaret,
   fullHeight = false,
+  assistiveText,
   ...marginProps
 }: FieldProps) => {
   return (
     <Container className={className} {...marginProps}>
       {label && (
         <Box mb={{ custom: outlined ? 4 : 0 }}>
```
mickaelzhang, line 42: Should we move this to the version with `renderAsTitle = false`? We already have some margin bottom when displayed as title

src/Field/Field.tsx
```diff
@@ -41,9 +41,17 @@ export const Field = ({
...
           {renderAsTitle ? (
-            <Text tag="label" typo="heading-small" htmlFor={id} mb="16px">
-              {label}
-            </Text>
+            <Box mb="16px">
+              <Text tag="label" typo="heading-small" htmlFor={id}>
+                {label}
+              </Text>
+
+              {assistiveText && (
```
fleonard, line 49: wondering if we should link the two prop from a type point of view?
Something like that if we have `renderAsTitle` then `assistiveText` is optional but available while if `renderAsTitle` is not set we can't even try to set `assistiveText`

src/Field/Field.tsx
Jmuskett: resolved in [82e535](https://github.com/marshmallow-insurance/smores-react/pull/2118/commits/82e535d107f5efa170c85b575cc570d45c8c77aa)

src/Field/Field.tsx
Jmuskett: resolved in [82e535](https://github.com/marshmallow-insurance/smores-react/pull/2118/commits/82e535d107f5efa170c85b575cc570d45c8c77aa)

src/TextInput/TextInput.stories.tsx
Jmuskett: resolved in [82e535](https://github.com/marshmallow-insurance/smores-react/pull/2118/commits/82e535d107f5efa170c85b575cc570d45c8c77aa)

src/Field/Field.tsx
Jmuskett: I've been looking at this this morning and I don't think this is possible given the way our types are constructed. We have interfaces to define the CommonFieldTypes, which are already extending from MarginProps. We'd need to create a discriminated union  like 

```
type AssistiveTextProps = 
| {renderAsTitle: true, assistiveText?: string}
| {renderAsTitle: false, assistiveText?: never}
```

but since the `CommonFieldTypes` is an interface, we can't then extend from a type.

If you have a different idea for how to do it, let me know and I'll tak ea look.

fleonard: If you check the TextInput component we have the exact same scenario?
```
interface Props extends CommonFieldTypes { ... }

type InputProps =
  | {
      /** on change is required and on input optional */
      onChange: (e: string) => void
      onInputChange?: (e: FormEvent<HTMLInputElement>) => void
    }
  | {
      /** on input is required and on change optional */
      onChange?: (e: string) => void
      onInputChange: (e: FormEvent<HTMLInputElement>) => void
    }

export type TextInputProps = Props & InputProps
```

fleonard: Or the problem is that `renderAsTitle` is part of `CommonFieldTypes`?

fleonard: I suppose we could convert everything to types and use type union instead of interfaces? At the end of the day we'll never have to change those types after they are created since it's a library so we can drop interfaces?
@mickaelzhang what do you think?

Jmuskett: > Or the problem is that `renderAsTitle` is part of `CommonFieldTypes`?

yup, exactly this

Jmuskett: > 

also, if we do decide to do this, I'll do it as a separate PR.

src/Field/Field.tsx
```diff
@@ -32,28 +33,45 @@ export const Field = ({
...
-              {label}
-              {required && (
-                <Text tag="span" typo="body-small" color="error">
-                  *
+            <Box mb="16px">
+              <Text
+                tag="label"
+                typo="heading-small"
+                htmlFor={id}
+                mb={{ custom: 4 }}
```
fleonard, line 49: should this be `4` only if we have `assistiveText`? cause we already have the whole wrapper with `mb="16px"` so at that point we should probably put a `mt` to `assistiveText` instead?

Jmuskett: Yup goof point, fixed in [482a4](https://github.com/marshmallow-insurance/smores-react/pull/2118/commits/482a475fa27e58e0c51bd62be2c09b971e3c8f78)

fleonard (approved):

PR #2148, opened 2022-08-18
Add OnBlur to SearchInput

fleonard:
## Screenshot / video

## What does this do?

- Give the reviewers context

### DIFF
--- a/src/SearchInput/SearchInput.tsx
+++ b/src/SearchInput/SearchInput.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useState } from 'react'
+import React, { FC, FormEvent, useState } from 'react'
 import styled from 'styled-components'
 import { darken } from 'polished'
 
@@ -19,6 +19,7 @@ export interface SearchInputProps extends CommonFieldTypes {
   onFound: (element: string) => void
   resultsRelativePosition?: boolean
   showIcon?: boolean
+  onBlur?: (e: FormEvent<HTMLInputElement>) => void
 }
 
 export const SearchInput: FC<SearchInputProps> = ({
@@ -32,6 +33,7 @@ export const SearchInput: FC<SearchInputProps> = ({
   outlined = false,
   showIcon = false,
   renderAsTitle = false,
+  onBlur,
   ...otherProps
 }) => {
   const id = useUniqueId(idProp)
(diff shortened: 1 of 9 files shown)


src/SearchInput/SearchInput.tsx
```diff
@@ -19,6 +19,7 @@ export interface SearchInputProps extends CommonFieldTypes {
   onFound: (element: string) => void
   resultsRelativePosition?: boolean
   showIcon?: boolean
+  onBlur?: (e: FormEvent<HTMLInputElement>) => void
```
mickaelzhang, line 22: I think the right typing would be:
```suggestion
  onBlur?: (e: FocusEvent<HTMLInputElement>) => void
```

src/SearchInput/SearchInput.tsx
```diff
@@ -92,6 +94,9 @@ export const SearchInput: FC<SearchInputProps> = ({
             onChange={updateInputState}
             outlined={outlined}
             selected={selected}
+            onBlur={(e) => {
+              onBlur && onBlur(e)
+            }}
```
mickaelzhang, line 99:
```suggestion
            onBlur={onBlur}
```

src/SearchInput/SearchInput.tsx
fleonard: We are using form event everywhere else tho 🤔

fleonard: Should I change it across all instances?

fleonard: MDN actually says it's a FocusEvent so I'll update it

githab666 (approved):

src/SearchInput/SearchInput.tsx
mickaelzhang: The typing is probably wrong yes, I believe for `onFocus` and `onBlur`, it should be a `FocusEvent`

mickaelzhang (dismissed):

fleonard:
@mickaelzhang I've realised I've never updated the Text Readme with the new typo so I've done it now

mickaelzhang (approved):
@fleonard I'm wondering if it is used? I never needed it on my end

PR #2185, opened 2022-08-31
Clean Field component

mickaelzhang:
## What does this do?

No UI changes but cleaned up some of the props that were doing nothing

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -64,7 +64,6 @@ export const Dropdown = forwardRef(function Dropdown(
   ref: ForwardedRef<HTMLSelectElement>,
 ) {
   const id = useUniqueId(idProp)
-  const [key, setKey] = useState('')
   const [hasOptGroups, setHasOptGroups] = useState(false)
   const [dropdownItemsGroups, setDropdownItemsGroups] = useState(
     [] as DropdownItem[][],
@@ -74,9 +73,6 @@ export const Dropdown = forwardRef(function Dropdown(
     if (list.length === 1) {
       onSelect(list[0].value)
     }
-
-    // update 'key' props in order to deselect the active option if a new list passed
-    list.length > 1 ? setKey(`${list[0].value}-${list.length}`) : ''
 
     setHasOptGroups(!!list.find((item) => !!item.optionGroupLabel))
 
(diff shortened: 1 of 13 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -21,7 +21,6 @@ export type DropdownItem = {
 }
 
 export interface Props extends CommonFieldTypes {
-  ref?: RefObject<HTMLSelectElement>
```
mickaelzhang, line 24: This is useless

src/Dropdown/Dropdown.tsx
```diff
@@ -74,9 +72,6 @@ export const Dropdown = forwardRef(function Dropdown(
       onSelect(list[0].value)
     }
 
-    // update 'key' props in order to deselect the active option if a new list passed
-    list.length > 1 ? setKey(`${list[0].value}-${list.length}`) : ''
```
mickaelzhang, line 78: This was a fix from 3 years ago but couldn't reproduce
This seems more like an hack than a real fix so I was thinking of removing it for now

src/Field/Field.tsx
```diff
@@ -72,21 +59,7 @@ export const Field = ({
...
-        outlined={outlined}
-        error={error}
-        key={dropdownKey ?? null}
-      >
-        {children}
-        {showCaret && (
-          <Caret outlined={outlined}>
-            <Icon render="caret" color="subtext" size={24} />
-          </Caret>
-        )}
```
mickaelzhang, line 87: Caret has been moved to Dropdown since it's only used there

src/Field/Field.tsx
```diff
@@ -72,21 +59,7 @@ export const Field = ({
         </>
       )}
 
-      <Content
-        fullHeight={fullHeight}
-        value={value}
-        outlined={outlined}
-        error={error}
-        key={dropdownKey ?? null}
```
mickaelzhang, line 80: Checked and none of the props were doing anything

src/Fieldset/Fieldset.tsx
```diff
@@ -1,11 +1,11 @@
-import React from 'react'
+import React, { ReactNode } from 'react'
 
 import { MarginProps } from '../utils/space'
 import { Box } from '../Box'
 import { Text } from '../Text'
 
 type FieldSetProps = {
-  children: React.ReactChild
+  children: ReactNode
```
mickaelzhang, line 8: Update to be able to be more flexible in what is passed as children

Jmuskett (approved):

PR #2186, opened 2022-08-31
Setup `RadioGroup` structure

mickaelzhang:
## Screenshot / video

Screenshot 2022-08-31 at 16 14 22

## What does this do?

Still in WIP but can be merged for now as it isn't exported, works that still needs to be done:
- Use `Fieldset` with some update around label
- Add error state
- Implement `displayType`: `vertical-card` and `horizontal-card`

Those will be done in the next few PRs

### DIFF
--- a/src/RadioGroup/RadioElement.tsx
+++ b/src/RadioGroup/RadioElement.tsx
@@ -1,0 +1,63 @@
+import React, { FC } from 'react'
+import styled from 'styled-components'
+import { MarginProps } from '../utils/space'
+import { theme } from '../theme'
+import { Box } from '../Box'
+import { visuallyHidden } from '../utils/visuallyHidden'
+import { focusOutline } from '../utils/focusOutline'
+
+type RadioElementProps = {
+  name: string
+  id: string
+  value: string
+  checked: boolean
+  onChange: (value: string) => void
+} & MarginProps
+
+export const RadioElement: FC<RadioElementProps> = ({
+  name,
+  id,
+  value,
+  checked,
+  onChange,
+  ...marginProps
+}) => {
+  return (
+    <>
+      <StyledInput
+        type="radio"
+        name={name}
+        id={id}
+        value={value}
+        checked={checked}
+        onChange={() => onChange(value)}
+      />
+      <RadioCircle {...marginProps} />
+    </>
+  )
+}
+
+const StyledInput = styled.input`
+  ${visuallyHidden}
+`
+
+const RADIO_SIZE = 24
+
+const RadioCircle = styled(Box)`
+  flex-shrink: 0;
+  width: ${RADIO_SIZE}px;
+  height: ${RADIO_SIZE}px;
+  border-radius: ${RADIO_SIZE}px;
+  background-color: ${theme.colors.white};
+  border: 2px solid ${theme.colors.subtext};
+
+  &:hover {
+    border: 2px solid ${theme.colors.secondary};
+  }
+
+  ${StyledInput}:checked + & {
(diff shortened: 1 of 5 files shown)


src/RadioGroup/RadioItem.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { FC } from 'react'
+import { useUniqueId } from '../utils/id'
+import { RadioElement } from './RadioElement'
+import styled from 'styled-components'
+import { theme } from '../theme'
```
ellajay, line 5: Does S'mores follow the pattern of separating imports according to external / internal components / utils? It's been a while since I've worked on this repo so can't remember

src/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+  onChange: (value: string) => void
+  value: string
+}
+
+export const RadioGroup: FC<RadioGroupProps> = ({
+  options,
+  onChange,
+  value,
+}) => {
+  const name = useUniqueId()
```
ellajay, line 20: Why is a unique id set as the name here then passed down to the `RadioItem` and `RadioElement` - how is the name used?

src/RadioGroup/RadioElement.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  &:hover {
+    border: 2px solid ${theme.colors.secondary};
+  }
+
+  ${StyledInput}:checked + & {
+    border: 8px solid ${theme.colors.secondary};
+  }
+
+  ${focusOutline({ selector: `${StyledInput}:focus-visible + &` })}
```
ellajay, line 62: What does the `+ & `do here?

src/utils/visuallyHidden.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+export const visuallyHidden = `
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+`
```
Jmuskett, line 13: nice

Jmuskett (approved):

src/RadioGroup/RadioItem.tsx
mickaelzhang: I think we do, I always forgot to follow what we agreed on tbh
I'll make an update

src/RadioGroup/RadioGroup.tsx
mickaelzhang: `useUniqueId` is an utility to generate an unique identifier, it's not related to the `id` prop per say even if it's the most common usage

Here we generate an unique identifier for the `name` prop on each option so that they are linked together

src/RadioGroup/RadioElement.tsx
mickaelzhang: Here's the breakdown of what is does
- `${StyledInput}:focus-visible`: `StyledInput`  is the checkbox element and we apply style when the `focus-visible` state is active
- `+` is for the adjacent/next sibling element, so in this case what is next to `StyledInput`
- `&` is a `styled-component` feature allowing us to refer to the current declared element (`RadioCircle`)

So what this selector does is apply CSS on `RadioCircle` when the checkbox is `focus-visible`

ellajay: Amazing, thanks for the clear breakdown 😊

ellajay (approved):

PR #2188, opened 2022-09-01
Fields update

mickaelzhang:
## Screenshot / video
Screenshot 2022-09-01 at 11 42 26
Screenshot 2022-09-01 at 11 42 37

## What does this do?

- Add Fieldset on RadioGroup
- Share UI/logic for Field and Fieldset


src/fields/components/InternalField.tsx
```diff
@@ -1,52 +1,59 @@
...
+interface InternalFieldProps extends InternalCommonFieldProps {
   children: ReactNode
+  className?: string
   assistiveText?: string
+  htmlFor?: string
+  fieldType?: 'field' | 'fieldset'
 }
 
-export const Field = ({
+export const InternalField = ({
```
mickaelzhang, line 16: Created an internal component that can be used on both `Field` and `Fieldset` so now they share the same UI and logic

src/fields/components/Placeholder.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { Box } from '../../Box'
+import { theme } from '../../theme'
+
+export const Placeholder = () => {
```
mickaelzhang, line 7: Screenshot 2022-09-01 at 11 34 00

Placeholder component for stories

src/fields/components/InternalField.tsx
```diff
@@ -70,9 +77,12 @@ export const Field = ({
...
 
-const Container = styled(Box)<{ className: string }>`
+const Container = styled(Box)`
   display: flex;
   flex-direction: column;
   position: relative;
   width: 100%;
+
+  // In case, the element is a 'fieldset', we remove the border
+  border: 0;
```
mickaelzhang, line 87: @Jmuskett This should fix the issue

src/fields/components/InternalField.tsx
```diff
@@ -1,52 +1,59 @@
...
+      className={className}
+      {...marginProps}
+    >
       {label && (
         <>
           {renderAsTitle ? (
             <Box mb="16px">
               <Text
-                tag="label"
+                tag={fieldType === 'field' ? 'label' : 'legend'}
```
mickaelzhang, line 41: Only difference between the 2 are the tags rendered for a11y purpose

src/fields/commonFieldTypes.ts
```diff
@@ -0,0 +1,18 @@
...
+  label?: string
+  renderAsTitle?: boolean
+  assistiveText?: string
+  required?: boolean
+  error?: boolean
+  errorMsg?: string
+}
+
+export interface InternalCommonFieldProps extends CommonFieldProps {
+  outlined?: boolean
```
mickaelzhang, line 17: `outlined` isn't needed on all component so I've moved it

src/NumberInput/NumberInput.tsx
```diff
@@ -10,12 +10,12 @@ import styled from 'styled-components'
...
 import { Icon } from '../Icon'
-import { Field, CommonFieldTypes } from '../Field'
+import { Field, CommonFieldProps } from '../fields/Field'
 
 import { theme } from '../theme'
 import { useUniqueId } from '../utils/id'
 
-export interface Props extends CommonFieldTypes {
+export interface Props extends CommonFieldProps {
   type?: 'number' | 'tel'
```
githab666, line 19: Not sure `'number' and 'tel'` are very clear, could we change to something more helpful?

src/fields/components/InternalField.tsx
githab666: Love this

githab666 (approved):
Minor comments - looking good :shipit:

src/NumberInput/NumberInput.tsx
mickaelzhang: This is part of the API for `NumberInput` so changing might be breaking some stuff
Better keep it as it is for now but might be worth to have a look later on 👍

Jmuskett (approved):

PR #2189, opened 2022-09-01
Add RadioGroup UI variant

mickaelzhang:
## Screenshot / video

Screenshot 2022-09-01 at 16 56 56
Screenshot 2022-09-01 at 16 57 03
Screenshot 2022-09-01 at 16 57 12
Screenshot 2022-09-01 at 16 57 17

## What does this do?

Still in WIP but should be useable now


src/RadioGroup/RadioElement.tsx
```diff
@@ -22,8 +23,13 @@ export const RadioElement: FC<RadioElementProps> = ({
   value,
   checked,
   onChange,
+  isError,
   ...marginProps
 }) => {
+  console.log({
```
githab666, line 29: Is there a reason this has been left in?

githab666 (dismissed):
Just one minor comment - looks great! :shipit:

src/RadioGroup/RadioElement.tsx
mickaelzhang: Nop, forgot about it, will remove thanks!

fleonard (approved):
LGTM

PR #2202, opened 2022-09-06
Improve typing for `RadioGroup` value

mickaelzhang:
## What does this do?

Improved typing for the prop `value`, it should now infer what is the correct type and can be more precise than `string` when possible. The type of `value` is shared and used for both the prop `value` and `onChange`

- When the type is only `string`
Screenshot 2022-09-06 at 11 33 13

- Using `boolean` as value
Screenshot 2022-09-06 at 11 32 45

- Having a subset of `string`
Screenshot 2022-09-06 at 11 33 32

### DIFF
--- a/src/RadioGroup/RadioElement.tsx
+++ b/src/RadioGroup/RadioElement.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react'
+import React from 'react'
 import styled from 'styled-components'
 
 import { theme } from '../theme'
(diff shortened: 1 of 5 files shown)


src/RadioGroup/RadioElement.tsx
```diff
@@ -7,32 +7,33 @@ import { visuallyHidden } from '../utils/visuallyHidden'
...
-  value: string
+  value: Value
   checked: boolean
-  onChange: (value: string) => void
+  onChange: (value: Value) => void
   isError: boolean
 } & MarginProps
 
-export const RadioElement: FC<RadioElementProps> = ({
+export const RadioElement = <Value extends BaseValueType>({
```
mickaelzhang, line 21: We can't use `FC` when using generic

fleonard (approved):

PR #2224, opened 2022-09-12
SearchInput cleanup

mickaelzhang:
## What does this do?

No behaviour change, just cleaning up `SearchInput` so that I can make it work when I'll add the `value` prop in my following PR


fleonard (approved):
Nice cleanup!

PR #2253, opened 2022-09-21
Fix Dropdown's `defaultValue`/`value` props behaviour

mickaelzhang:
## Screenshot / video

This message is showing up when `value` is provided
Screenshot 2022-09-21 at 15 45 34

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -14,6 +14,7 @@ import { Icon } from '../Icon'
 import { Field } from '../fields/Field'
 import { CommonFieldProps } from '../fields/commonFieldTypes'
 import { useUniqueId } from '../utils/id'
+import { useControllableState } from '../utils/useControlledState'
 
 export type DropdownItem = {
   optionGroupLabel?: string
@@ -50,7 +51,7 @@ export const Dropdown = forwardRef(function Dropdown(
     id: idProp,
     placeholder,
     name,
-    value,
+    value: valueProp,
     defaultValue,
     disabled = false,
     list,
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -63,6 +64,10 @@ export const Dropdown = forwardRef(function Dropdown(
   }: DropdownProps,
   ref: ForwardedRef<HTMLSelectElement>,
 ) {
+  const [value, setValue] = useControllableState({
+    initialState: defaultValue,
+    stateProp: valueProp,
+  })
```
mickaelzhang, line 70: `Dropdown` behaviour when used is still the same
Internally tho, it's always controlled even when only `defaultValue` is provided

src/Dropdown/Dropdown.tsx
```diff
@@ -94,17 +99,13 @@ export const Dropdown = forwardRef(function Dropdown(
...
-          }
           disabled={disabled || list.length < 1}
-          onChange={(e: FormEvent<HTMLSelectElement>) => {
-            onSelect && onSelect(e.currentTarget.value)
-            onInputChange && onInputChange(e)
+          onChange={(event) => {
+            const value = event.currentTarget.value
+
+            onSelect?.(value)
+            onInputChange?.(event)
```
mickaelzhang, line 107: Not sure a second event is needed tbh, is it used?

src/Dropdown/Dropdown.tsx
```diff
@@ -198,23 +199,16 @@ const StyledSelect = styled.select<UsesOutline>`
...
-  &:focus-within {
-    border: 2px solid ${theme.colors.outline};
-  }
-`}
-  ${({ value }) =>
-    value &&
-    value != '' &&
-    `
-    border: 2px solid ${theme.colors.outline};
-`}
```
mickaelzhang, line 216: I've removed this, I think it was not intended, there is a border when the component is controlled

src/Dropdown/Dropdown.tsx
fleonard: yeah I think FNOL added it cause `onSelect` return the string value while they needed the actual event

fleonard (approved):

src/Dropdown/Dropdown.tsx
josianed: `onInputChange` is not used at all throughout FNOL. It was added before the correct use of the input components & react-hook-forms was established and can be removed now

PR #2266, opened 2022-09-28
Remove Dropdown internal `useEffect`

mickaelzhang:


src/Dropdown/Dropdown.tsx
```diff
@@ -69,18 +68,9 @@ export const Dropdown = forwardRef(function Dropdown(
...
   const id = useUniqueId(idProp)
-  const [hasOptGroups, setHasOptGroups] = useState(false)
-  const [dropdownItemsGroups, setDropdownItemsGroups] = useState(
-    [] as DropdownItem[][],
-  )
-
-  useEffect(() => {
-    if (list.length === 1) {
-      onSelect(list[0].value)
-    }
```
mickaelzhang, line 80: This changed the behaviour of `Dropdown` slightly but I don't think this would be considered a breaking change?

fleonard: I don't think it's a massive deal if it doesn't auto select the element if there's only one.

fleonard (approved):

PR #2299, opened 2022-10-11
Fix searchInput when multiple items has the same value

fleonard:
## What does this do?

The SearchInput has an issue that if multiple items in the list has the same value when you select one of them it will always display the first of the list as label.

e.g. for countries `Britain`, `UK` and `United Kingdom` all has the same option value as `GB`. If you search and select `UK` the searchInput will always display `Britain` because it's the first label that he finds with `GB` value.

This PR fixes that so if you search for `UK`, `UK` will be displayed in the input. It will still select `Britain` for any persisted value that we pass to `SearchInput` because we essentially lose the label associated to the value the we pass down as a prop and will fallback to the first one he finds in the search list. Not really sure if there's a real fix for it unless we find a way to pass down both label and values to the component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## [2.8.5] - 2022-10-10
-## Added
+## [2.9.0] - 2022-10-11
+### Changed
+- Update SearchInput logic to select correct item when value is the same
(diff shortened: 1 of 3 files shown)


githab666 (approved):
:shipit:

PR #2304, opened 2022-10-13
make radio button container clickable

liampuk:
## What does this do?

Makes the container of the radio button clickable, instead of just the radio button element.

### DIFF
--- a/src/ConfirmationRadioButtons/RadioButtonStyled.tsx
+++ b/src/ConfirmationRadioButtons/RadioButtonStyled.tsx
@@ -1,7 +1,6 @@
 import React, { FC, ChangeEvent, FocusEvent, ReactElement } from 'react'
 import styled from 'styled-components'
 import { theme } from '../theme'
-import { Box } from '../Box'
 
 export interface FakeInput {
   checked?: boolean
@@ -28,7 +27,7 @@ export const RadioButton: FC<RadioButtonProps> = ({
   onBlur,
   disabled = false,
 }: RadioButtonProps) => (
-  <Box flex alignItems="center">
+  <InputContainer>
     <RadioInput
       id={id}
       type="radio"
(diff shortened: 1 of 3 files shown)


src/ConfirmationRadioButtons/RadioButtonStyled.tsx
```diff
@@ -28,7 +27,7 @@ export const RadioButton: FC<RadioButtonProps> = ({
   onBlur,
   disabled = false,
 }: RadioButtonProps) => (
-  <Box flex alignItems="center">
+  <InputContainer>
```
fleonard, line 30: instead of adding a new StyledComponent you can just use `as="label"`

fleonard (commented):
Considering that now we have a proper `RadioGroup` component I was wondering if might be better to replace all the Confirmation with the new component?

src/ConfirmationRadioButtons/RadioButtonStyled.tsx
fleonard: @liampuk no what I'm saying is that the S'mores `Confirmation` button should be deprecated in favour of [this](https://marshmallow-insurance.github.io/smores-react/?path=/story/radiogroup--horizontal)

liampuk: Yep that's what I mean, it's maybe worth adding a component that uses `RadioGroup` with Yes/No as that is used quite a lot in agent-portal/fnol?

fleonard: you just need to pass this as options to get a YES/NO component
```
options={[
   { label: 'Yes', value: true },
   { label: 'No', value: false },
]}
 ```
 we do it already in the new direct flow in signup.

PR #2351, opened 2022-11-03
Add fossa analysis step to CI

bspammer:
[As requested here](https://eatmarshmallows.slack.com/archives/C8J3F9PN1/p1667389007431219?thread_ts=1667385398.687619&cid=C8J3F9PN1), this should run the fossa tool on every merge to main. We need to install the fossa CLI, and yq, then we use the FOSSA_API_KEY organization secret to run the analysis and upload it.

### DIFF
--- a/.fossa.yml
+++ b/.fossa.yml
@@ -1,0 +1,8 @@
+version: 3
+
+project:
+  id: smores-react
+  name: smores-react
+  url: https://github.com/marshmallow-insurance/smores-react
+  team: frontend
+  policy: Frontend Security Policy
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -20,16 +20,9 @@ jobs:
           npm ci
           npm run build-storybook
       - name: Fossa dependency analysis 🐛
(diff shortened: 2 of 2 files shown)


fleonard:
@bspammer have you had a look at [these](https://github.com/fossas/fossa-action) by any chance? Wondering if we might just use their own actions?

bspammer:
@fleonard Nice! No I just assumed they hadn't published an action, that should simplify it somewhat.

fleonard:
@bspammer yeah it will also be much better cause if we just use the latest everything they update will be automatically picked up

bspammer:
@fleonard actually looking again this action doesn't have any way of specifying the team, nor of specifying the security policy. We would need to add `opslevel.yaml` to the repo to make this work. It's up to you what you'd prefer really.

fleonard:
@bspammer we already a `yaml` file for `opslevel` so shouldn't be an issue to add one for fossa

bspammer:
@fleonard Ok sure, it will make it more inconsistent with the other repos but I guess that ship has sailed already 😁

fleonard (approved):

PR #2354, opened 2022-11-04
Update to React 18

fleonard:
## What does this do?

Update S'mores to React 18 and tweak peer deps to make it available to older versions.


.storybook/main.js
```diff
@@ -1,9 +1,6 @@
 module.exports = {
   stories: ['../src/**/*.stories.tsx'],
-  addons: [
-    '@storybook/addon-essentials',
-    'storybook-addon-performance/register',
```
fleonard, line 5: Storybook performance doesn't yet support React 18 so I've removed it for now. We can add it back once they do the update

package.json
```diff
@@ -39,12 +39,11 @@
     "@storybook/react": "^6.2.6",
     "@storybook/storybook-deployer": "^2.8.7",
     "@testing-library/jest-dom": "^5.11.4",
-    "@testing-library/react": "^12.0.0",
-    "@testing-library/react-hooks": "^8.0.0",
```
fleonard, line 43: `@testing-library/react-hooks` is now deprecated since `@testing-library/react` ship it's own version of `renderHook`

mickaelzhang (commented):
Some component's doesn't needs `PropsWithChildren` like the `SupportMessage` or the `SearchInput` I think?

fleonard:
@mickaelzhang yeah you are right. I've actually used a script to fix the PropsWithChildren so it might have added it to components that might not need it. I'll double check

mickaelzhang (dismissed):

fleonard:
@mickaelzhang I've removed `PropsWithChildren` since we were already exporting `children` already and you were actually right that using it was complaining in the IDE

.github/CODEOWNERS
```diff
@@ -15,4 +15,4 @@
 # This should make it easy to add new rules without breaking existing ones.
 
 # Global - These owners will be the default owners for everything in the repo.
-*           @ellajay @fleonard @habin-isa @jmuskett @josianed @mickaelzhang
+*           @ellajay @fleonard @habin-isa @josianed @mickaelzhang
```
mickaelzhang, line 18: Wondering if maybe we could change this to the frontend group?

package-lock.json
```diff
@@ -127,9 +127,9 @@
       }
     },
     "node_modules/@babel/generator": {
-      "version": "7.20.2",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.20.2.tgz",
-      "integrity": "sha512-SD75PMIK6i9H8G/tfGvB4KKl4Nw6Ssos9nGgYwxbgyTP0iX/Z55DveoH86rmUB/YHTQQ+ZC0F7xxaY8l2OF44Q==",
+      "version": "7.20.4",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.20.4.tgz",
+      "integrity": "sha512-luCf7yk/cm7yab6CAW1aiFnmEfBJplb/JojV56MYEK7ziWfGmFlTfmL9Ehwfy4gFhbjBfWO1wj7/TuSbVNEEtA==",
```
mickaelzhang, line 132: Is it normal that the package-lock.json has changed on this PR without an update on `package.json`?

fleonard: no it's not I think the rebase messed everything up 😢 
I'll try to fix it

.github/CODEOWNERS
fleonard: the problem is that in the FE team there are a lot of BE and I want to avoid spamming them 😁

mickaelzhang (approved):

PR #2487, opened 2023-01-25
Add Responsive props

mickaelzhang:
## Screenshot / video

## What does this do?

- Give the reviewers context

### DIFF
--- a/src/utils/responsiveProp.ts
+++ b/src/utils/responsiveProp.ts
@@ -1,0 +1,47 @@
+const breakpointToMinPxMapping = {
+  small: 0,
+  medium: 768,
+  large: 1024,
+} satisfies Record<string, number>
+
+type Breakpoint = keyof typeof breakpointToMinPxMapping
+const breakpointList = Object.keys(breakpointToMinPxMapping) as Breakpoint[]
+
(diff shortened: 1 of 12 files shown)


src/utils/responsiveProp.ts
```diff
@@ -0,0 +1,47 @@
...
+    value !== null &&
+    breakpointList.includes(Object.keys(value)[0] as any)
+  )
+}
+
+export const resolveResponsiveProp = <PropValue>(
+  value: ResponsiveProp<PropValue>,
+  styleTemplate: (value: PropValue) => string,
+) => {
+  if (!isPropByBreakpoint(value)) {
```
fleonard, line 27: oh so this is what you were mentioning around having to set it in a media query anyway just to keep the correct order of the styles.

fleonard (approved):

src/utils/responsiveProp.ts
mickaelzhang: Yeah, you can try to remove the media query, it will change the behaviour

fleonard: So I guess we just need to be sure if we wanna add more breakdpoints to have the lowest one always at 0 for this edge case

PR #2522, opened 2023-02-09
feat(tooltip): new component

DeadEnglish:
## Screenshot / video

TBC (waiting on figma access for final design)

## What does this do?

A new component that provides a tooltip to the child component. The tooltip can display in all four directions (top, right, bottom, left) and the content display can be either a `String` value or `ReactNode`

This does not impact any other components as this is a brand new component.

- [Ticket](https://www.notion.so/getmarshmallow/Create-Smore-s-tooltip-5443d43b8d6a450a9bc4ebdd44152ff3?pvs=4) 
- [RFC](https://www.notion.so/getmarshmallow/Tooltip-466c358dc61c4f65881a0ff23950ad56#8f35735575e445d6a01c2338143d0eae)
- [Figma Design](https://www.figma.com/proto/ek2Jsn7pjiWZBdtuolNlX0/Claims-Management-System-(CMS)?page-id=1960%3A7118&node-id=2086%3A19886&viewport=642%2C427%2C0.14&scaling=min-zoom&starting-point-node-id=2210%3A65262)


src/Tooltip/Tooltip.styles.ts
```diff
@@ -0,0 +1,88 @@
+import styled, { css } from 'styled-components'
```
fleonard, line 1: don't have strong personal preferences but we tend to keep all styled component at the bottom of the actual `tsx` file. Considering that this is a public library might be good to keep the same standards

src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,38 @@
...
+  content: string | ReactNode
+  position: TooltipPosition
+  underline?: boolean
+}
+
+export const Tooltip: FC<TooltipProps> = ({
+  children,
+  title,
+  content,
+  position,
```
fleonard, line 19: should we default this to top since it's probably the most commonly used?

Also based on Figma layout proposal I'm wondering how we are going to handle these? I know that Alistair is off so might be better to wait for him or to ask in the design-system channel about clarification?

- The positioning of the tooltip relative to the page is determined by the amount of space it has available 
- Tooltip is agnostic to the layout of the page and it’s surrounding elements, but it’s best to accommodate it in the whitespace

src/Tooltip/Tooltip.styles.ts
DeadEnglish: Yep no worries! I showed it to Rowan as I knew it might be a bit strange moving having a seperate `styles.ts` file. Have moved them underneath the component now 😀

src/Tooltip/Tooltip.tsx
DeadEnglish: About to drop a message in design-system, will update when I get a response :)

src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,268 @@
...
+  }, [tipContainer, tooltipPosition])
+
+  useEffect(() => {
+    window.addEventListener('resize', handleTipViewport)
+    window.addEventListener('scroll', handleTipViewport)
+    return () => {
+      window.removeEventListener('resize', handleTipViewport)
+      window.removeEventListener('scroll', handleTipViewport)
+    }
+  }, [handleTipViewport])
```
fleonard, line 79: We have an `useEventListener` hook that you can use here instead

src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,268 @@
...
+type ArrowPosition = 'left' | 'right' | 'center' | 'top' | 'bottom'
+
+export interface TooltipProps {
+  children: ReactNode
+  title?: string
+  content: string | ReactNode
+  position: Position
+  underline?: boolean
+  size?: Size
+  arrowPosition?: ArrowPosition
```
fleonard, line 24: considering that we dynamically set the tooltip position based on the element, should we also derive the arrow position as well and remove both `position` and `arrowPosition` as props?
Just trying to avoid conflicting behaviours between default values and resizes. 
If we agree with design that the default position is always top, then I guess everything else will change based on scrolling and resize. 
I know flexibility is a good thing but sometimes too much might not be? What do you think?

src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,268 @@
...
+  useRef,
+  useState,
+} from 'react'
+import styled, { css } from 'styled-components'
+import { Text } from '../Text'
+import { theme } from '../theme'
+
+type Position = 'top' | 'bottom' | 'left' | 'right'
+type Size = 'small' | 'large'
+type ArrowPosition = 'left' | 'right' | 'center' | 'top' | 'bottom'
```
fleonard, line 15:
```suggestion
type ArrowPosition = Position | 'center' 
```

src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+export interface TooltipProps {
+  children: ReactNode
+  title?: string
+  content: string | ReactNode
+  position: Position
+  underline?: boolean
+  size?: Size
```
fleonard, line 23: can't we derive the `size` from the content and maybe have just a fixed `max-width`? In the Figma file doesn't seem like there's specified any specific values for the size

DeadEnglish: Ahh, you're right, I was running off the M5 Figma and not the design systems. Will add a max width of `240px` (largest tooltip used on the M5 design).

fleonard: Sweet!

src/Tooltip/Tooltip.tsx
DeadEnglish: @fleonard I've taken some time to have a look and I _think_ I've come up with a decent solution. From the Figma file the arrow position looks like it can (and should) be placed anywhere so I've left this to be used defineable as a prop (the default will still be `center`) and removed the `position` prop as we're defaulting the tooltip to always display at the top.

The Tooltip now checks the page bounds and makes sure it's in view, defaulting to top and moving around the content if out of bounds. I've also made it default back to top when the space is available. 

I think this allows enough flexibility with the arrow position whilst not being too complex in implementation due to removing the tooltip `position` prop. Let me know what you think!

src/Tooltip/Tooltip.tsx
```diff
@@ -7,84 +7,124 @@ import React, {
...
     element.top >= 0 &&
     element.left >= 0 &&
     element.bottom <= window.innerHeight &&
     element.right <= window.innerWidth
 
   const handleTipViewport = useCallback(() => {
+    const shouldChangeXAxis: boolean =
+      tooltipPosition === 'left' || tooltipPosition === 'right'
+    const shouldChangeYAxis: boolean =
+      tooltipPosition === 'top' || tooltipPosition === 'bottom'
```
fleonard, line 51:
```suggestion
    const shouldChangeXAxis =
      tooltipPosition === 'left' || tooltipPosition === 'right'
    const shouldChangeYAxis =
      tooltipPosition === 'top' || tooltipPosition === 'bottom'
```

the type should be inferred, right?

src/Tooltip/Tooltip.tsx
```diff
@@ -7,84 +7,124 @@ import React, {
...
+
+    const handleArrowAxisChange = (): void => {
```
fleonard, line 53:
```suggestion
    const handleArrowAxisChange = () => {
```
same here?

fleonard (commented):
couple of minor things but it's looking good!

src/Tooltip/Tooltip.tsx
DeadEnglish: Have updated now 😄

src/Tooltip/Tooltip.tsx
DeadEnglish: Have updated now 😄

fleonard (approved):

PR #2532, opened 2023-02-16
feat(accordion): defaultIsOpen prop for accordion status

DeadEnglish:
## Screenshot / video
Usage in action:
Screenshot 2023-02-16 at 11 00 04

Figma design for CMS
Screenshot 2023-02-16 at 11 00 45

## What does this do?

Adding the ability to set the default open/close status for the `Accordion` component due to CMS needing this component to be open by default. This shouldn't affect any current usage of the `Accordion` component as the default is set to `false`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [2.17.0] - 2023-02-16
+### Changed
+- Added optional default position to `Accordion` component
 ## [2.16.0] - 2023-02-13
 ### Changed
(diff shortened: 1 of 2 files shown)


fleonard:
@DeadEnglish bump changelog and it's good to go!

DeadEnglish:
@fleonard Realised as I committed, should be good now 😄

fleonard (approved):

PR #2534, opened 2023-02-16
fix: export tooltip in index.ts

DeadEnglish:
## What does this do?

Correctly exporting the `Tooltip` component in the index.ts file.


src/index.ts
```diff
@@ -31,6 +31,7 @@ export * from './Textarea'
 export * from './TextInput'
 export * from './theme'
 export * from './Toggle'
+export * from './Tooltip'
```
fleonard, line 34: ahah I always forget to do it as well

fleonard (approved):

PR #2539, opened 2023-02-20
properly centre modal

liampuk:
## What does this do?

- Wrap modal in portal to properly centre in all cases

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mrshmllw/smores-react",
-  "version": "2.17.1",
+  "version": "2.17.2",
   "main": "./dist/index.js",
   "description": "Collection of React components used by Marshmallow Technology",
   "keywords": [
--- a/src/Modal/Modal.stories.tsx
+++ b/src/Modal/Modal.stories.tsx
@@ -3,9 +3,6 @@ import { Button } from '../Button'
 import { Box } from '../Box'
 import { Modal, ModalProps } from './Modal'
 
-const placeholderText =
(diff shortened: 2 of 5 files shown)


src/Modal/Modal.stories.tsx
```diff
@@ -12,18 +12,6 @@ export default {
   argTypes: { handleClick: { action: 'Callback - `handleClick`' } },
 }
 
-export const Playground = Modal.bind({})
```
fleonard, line 15: Think you might need to delete `WithLongContent` as well

src/Modal/Modal.tsx
```diff
@@ -39,35 +40,38 @@ export const Modal: FC<ModalProps> = ({
...
-      drawer={drawer}
-      width={width || '460px'}
-      className={containerClass}
-    >
-      <Box flex alignItems="flex-start" justifyContent="space-between" mb="8px">
-        <TitleElements flex direction="column">
-          {icon !== '' && (
-            <Icon render={icon} size={42} color="secondary" mb="16px" />
+}) => {
+  return createPortal(
```
fleonard, line 44: should we add the click outside thing already? it's a one liner basically

fleonard (dismissed):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mrshmllw/smores-react",
-  "version": "2.17.1",
+  "version": "2.17.2",
```
fleonard, line 3: you might have to manually add the release notes if you pushed tags before commititng the changelog updates

fleonard (approved):

PR #2546, opened 2023-02-22
Allow modal portal container to be dynamic

fleonard:
## What does this do?

Signup has a couple of peculiar modals and I need to be able to decide where to append the modal in order to override some styles and I'd rather do it in two portals with a `ref` instead of having to add the specific class styled in globalStyles because the modal is appended at the end of the body


CHANGELOG.md
```diff
@@ -555,7 +558,8 @@
 ### Changed
 - Updated gap and styles on Row component
 
-[2.17.2]: https://github.com/marshmallow-insurance/smores-react/compare/v2.17.1...v2.16.2
```
liampuk, line 558: 🙃

liampuk (approved):

PR #2559, opened 2023-03-02
Modal animation

fleonard:
## Screenshot / video

https://user-images.githubusercontent.com/1053476/222508540-01fb0f64-4a6d-4e90-9e75-3354cb0e9857.mov

## What does this do?

- add new show/hide animation for Modal component

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactNode, useRef } from 'react'
+import React, { FC, ReactNode, useEffect, useRef, useState } from 'react'
 import styled, { css } from 'styled-components'
 import { theme } from '../theme'
 import { createPortal } from 'react-dom'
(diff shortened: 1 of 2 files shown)


src/Modal/Modal.tsx
```diff
@@ -45,13 +47,29 @@ export const Modal: FC<ModalProps> = ({
...
   const modalRef = useRef<HTMLDivElement>(null)
+  const hasOpened = useRef(false)
+  const [opened, setOpened] = useState(false)
 
   useBodyScrollLock({ node: modalRef.current, showModal })
 
+  useEffect(() => {
+    if (modalRef.current === null) return
+
+    if (!hasOpened.current && showModal) {
```
fleonard, line 58: `showModal` in the end is needed because we want to trigger it only on the first click, without it it will append it at first render

DeadEnglish: 😢

DeadEnglish (approved):

DeadEnglish:
May need to merge my bump PR and tags before this 😅

fleonard:
@DeadEnglish na it's fine to merge this first and then your bump. I've added my changes to the same tag you want to bump

PR #2637, opened 2023-04-05
GRW2716 - success state

matdiak:
## Screenshot / video

image

## What does this do?

Allows components using InternalField to render complete status

### DIFF
--- a/src/fields/commonFieldTypes.ts
+++ b/src/fields/commonFieldTypes.ts
@@ -11,6 +11,8 @@ export interface CommonFieldProps extends MarginProps {
   required?: boolean
   error?: boolean
   errorMsg?: string
+  showStatus?: boolean
+  status?: 'complete' | null
 }
 
 export interface InternalCommonFieldProps extends CommonFieldProps {
--- a/src/fields/components/InternalField.tsx
+++ b/src/fields/components/InternalField.tsx
@@ -1,9 +1,10 @@
 import React, { ReactNode } from 'react'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 
 import { Text } from '../../Text'
 import { Box } from '../../Box'
 import { InternalCommonFieldProps } from '../commonFieldTypes'
+import { Icon } from '../../Icon'
 
 interface InternalFieldProps extends InternalCommonFieldProps {
   children: ReactNode
@@ -25,9 +26,12 @@ export const InternalField = ({
   error,
   errorMsg,
   required,
+  status,
+  showStatus,
   ...marginProps
 }: InternalFieldProps) => {
   const labelTag = fieldType === 'field' ? 'label' : 'legend'
+
   return (
     <Container
       as={fieldType === 'field' ? 'div' : 'fieldset'}
@@ -74,9 +78,34 @@ export const InternalField = ({
           {errorMsg}
         </Text>
       )}
+
+      {/* When status is null, a container is rendered to avoid layout shift */}
+      {!(error && errorMsg) && showStatus && status !== undefined && (
+        <StatusWrapper displayStatus={!!status} mt={'8px'}>
+          <Icon render="included" size={16} color="success" />
+          <Text typo="caption" color="success">
+            Complete
+          </Text>
+        </StatusWrapper>
+      )}
     </Container>
   )
 }
+
+const StatusWrapper = styled(Box)<{ displayStatus: boolean }>`
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  width: 0;
+  overflow: hidden;
+
+  ${({ displayStatus }) =>
+    displayStatus &&
+    css`
+      transition: width 0.6s ease-in;
+      width: 100%;
+    `}
+`
 
 const Container = styled(Box)`
   display: flex;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [2.18.5] - UNRELEASED
+### Added
+- InternalField takes showStatus and status props
+- Internal field renders complete status
+
 ## [2.18.4] - 2023-03-28
 ### Changed
 - Set the z-index on the snackbar container
--- a/src/fields/Fieldset/Fieldset.stories.tsx
+++ b/src/fields/Fieldset/Fieldset.stories.tsx
@@ -39,3 +39,11 @@ WithSmallLabel.args = {
   ...defaultArgs,
   renderAsTitle: false,
 }
+
+export const WithStatus = Template.bind({})
+
+WithStatus.args = {
+  ...defaultArgs,
+  showStatus: true,
+  status: 'complete',
+}


src/fields/commonFieldTypes.ts
```diff
@@ -11,6 +11,8 @@ export interface CommonFieldProps extends MarginProps {
   required?: boolean
   error?: boolean
   errorMsg?: string
+  showStatus?: boolean
+  status?: 'complete' | null
```
matdiak, line 15: This approach will let us have different statuses without having to adjust the previous ones. We would potentially need to create an object to store the styles for each status if we decide to add more. 

Just to clarify. 
`showStatus` displays an empty space so that there is no layout shift it appears
'status' defines the status and when it's null there is just space

matdiak: We could possibly simplify this and make it complete specific. As currently you can pass it any status and it would render `Complete`. What do you think?

src/fields/components/InternalField.tsx
```diff
@@ -13,6 +15,16 @@ interface InternalFieldProps extends InternalCommonFieldProps {
...
 }
 
+export const statusContents: {
+  [key in Status]: {
+    icon: string
+    color: Color
+    textContent: string
+  }
+} = {
+  complete: { icon: 'included', color: 'success', textContent: 'Complete' },
```
matdiak, line 25: We discussed possibly adding another status so I took the approach of adding a flexible status prop. We could even potentially deprecate error in the future and include in this. 

We can also simplify all this and just make the prop display complete status by default.

src/fields/commonFieldTypes.ts
```diff
@@ -11,8 +11,14 @@ export interface CommonFieldProps extends MarginProps {
   required?: boolean
   error?: boolean
   errorMsg?: string
+  status?: {
+    type: Status
+    showStatus: boolean
+  }
```
fleonard, line 17: I think this is a bit overly complicated?
the `showStatus` is not needed cause it can be derived by the optional `status` property and at that point there's no real need of having a `type` you can just have `status?: 'complete' | ...`

src/fields/components/InternalField.tsx
fleonard: What other status other than complete we might wanna have?
Deprecating error might take quite a lot of time considering how widely uses is and the need of a dynamic error message might not fit with the way status has been implemented so I'm wondering if it's actually worth it going down this route?

fleonard (commented):
@matdiak I think you will also need to expose to all the other form items?

src/fields/commonFieldTypes.ts
matdiak: So it's made that way to avoid layout shift and allow for animation. 
If `status` is truthy everything renders so that there is additional space for the status
And the showStatus actually triggers the animation. 

We could have it as simple as `status?: 'complete' | ...` but then is there a way to animate it / make sure there is no layout shift each time the status comes in?

src/fields/components/InternalField.tsx
matdiak: So the designs considered adding incomplete status but we decided not to include it for now.

src/fields/components/InternalField.tsx
```diff
@@ -74,10 +88,49 @@ export const InternalField = ({
...
+          isError={!!(error && errorMsg)}
+          mt="8px"
+        >
+          <Icon
+            render={statusContents[status.type].icon}
+            size={16}
+            color={statusContents[status.type].color}
+          />
+          <Text typo="caption" color="success">
+            Complete
```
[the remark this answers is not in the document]
matdiak: Yes, it should

src/fields/commonFieldTypes.ts
fleonard: we have layout shift with errors as well right now and I don't think it's ever been an issue. It's impossible to pre-render something that doesn't exist cause the other issue is that if it's alway there you have an extra gap of white space that's a problem as much as the layout shift

fleonard: you can potentially try to always render it and play around with opacity and visibility in order to make it animate avoiding the white space gap

src/SearchInput/SearchInput.tsx
```diff
@@ -136,7 +136,13 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
             onChange={handleInputChange}
             outlined={outlined}
             selected={isSelected}
-            onBlur={onBlur}
+            onBlur={(e) => {
+              if (displayedInputText === '') {
```
matdiak, line 140: This sets the value back to selectedValue in case the user deleted the input and clicked away

src/SearchInput/SearchOptions.tsx
```diff
@@ -21,22 +22,24 @@ export const SearchOptions: FC<SearchOptionsProps> = ({
...
-            <li key={i} onClick={() => onSelect(el)}>
-              {el.label}
-            </li>
-          ))
-        ) : (
-          <li>No results</li>
-        )}
-      </ResultsList>
-    </StyledResultsContainer>
+    <BoxWithPositionRelative>
```
matdiak, line 25: Wrapping the search results in position relative to fix #2.

matdiak, line 25: Bug no 1 - wrapping results in position relative

src/SearchInput/SearchOptions.tsx
```diff
@@ -47,7 +50,7 @@ const StyledResultsContainer = styled.div<
   overflow-y: hidden;
   ${({ positionRelative }) => !positionRelative && 'position: absolute;'}
   width: 100%;
-  ${({ outlined }) => outlined && 'left: 0px; top: 90%;'};
+  ${({ outlined }) => outlined && 'left: 0px; top: -8px;'};
```
matdiak, line 53: Bug no 1
Positioning outlined results according to the new wrapper

src/fields/components/InternalField.tsx
```diff
@@ -25,9 +26,11 @@ export const InternalField = ({
   error,
   errorMsg,
   required,
+  showCompletedStatus,
```
fleonard, line 29: can we just call this `completed`?

src/fields/components/InternalField.tsx
```diff
@@ -74,10 +77,45 @@ export const InternalField = ({
           {errorMsg}
         </Text>
       )}
+
+      {showCompletedStatus !== undefined && (
```
fleonard, line 81: is there any specific need for checking undefined? cause this will render the whole thing also if `false` while I suppose we just need it in case it's true?

src/fields/components/InternalField.tsx
matdiak: No problem

src/fields/components/InternalField.tsx
matdiak: So having spoken to Luke, out of all options- he prefers that there is white space underneath each input. 

So for `showCompletedStatus`
`undefined` => no whitespace
`false` => whitespace
`true` => complete status

matdiak: I did play with other options and I think it is possible to have it extend in height and width but then Luke still preferred whitespace. 

Is your main idea of using scale so that it does extend in height or something else?

fleonard: Yeah so the problem is that if you animate the `height` property to zero the margin top of the wrapper will still be rendered and that's what's gonna add the white space.
If you use `scale(0)` it will also include the margin so you end up having a `0 x 0` size element

matdiak: I see - I didn't know that. We can still chat with Luke if you think having it animated on y-axis would be better than whitesapce

fleonard: I'm not sure if there's another option of maybe moving the margin top to a wrapper inside the one we are animating, maybe worth a try?
So you essentially use `StatusWrapper` just for handling the display animation and then you wrap what's inside with a `Box` that handled the margin

matdiak: @leo, I just chatted to Luke and we tested the success state locally. He is happy the way it looks.

Just to sum up this conversation about the height. I understood that you are mentioning scale only in regards to animating it vertically. However, I see that what you mean is that when we have an error, despite the status having 0 height, there is still 8px margin applied to it.

So just to summarise
- we don't want margin (from status) when there is an error

Opt 1 - We set margin based on error 
```
{completed !== undefined && (
        <StatusWrapper
          displayStatus={completed}
          isError={!!(error && errorMsg)}
          mt={error && errorMsg ? '0' : '8px'} //HERE
        >
          <Icon render="included" size={16} color="success" />
          <Text typo="caption" color="success">
            Complete
          </Text>
        </StatusWrapper>
      )}
```

Opt 2 - We wrap the contents in a Box that sets margin
- AnimationWrapper does the animating
- StatusWrapper sets

```
{completed !== undefined && (
        <AnimationWrapper
          displayStatus={completed}
          isError={!!(error && errorMsg)}
        >
          <StatusWrapper mt={'8px'}>
              <Icon render="included" size={16} color="success" />
              <Text typo="caption" color="success">
                Complete
              </Text>
          </StatusWrapper>
        </AnimationWrapper>
      )}
```

```
const AnimationWrapper = styled(Box)<{
  displayStatus: boolean
  isError: boolean
}>`
  width: 0;
  overflow: hidden;

  ${({ displayStatus }) =>
    displayStatus &&
    css`
      transition: width 0.6s ease-in;
      width: 100%;
    `}

  /* This enables animation to appear when previous state is error */
  ${({ isError }) =>
    isError &&
    css`
      height: 0px;
    `}
`

const StatusWrapper = styled(Box)`
  display: flex;
  align-items: center;
  gap: 4px;
`
```

Which one do you prefer? I would actually lean towards the Opt 2. 

Btw `transform: scale(0)` [does not work for margin](https://stackoverflow.com/questions/48682497/why-does-the-scale-css-transformation-function-not-affect-margin).

fleonard: yeah I think we can go with option 2

src/fields/components/InternalField.tsx
```diff
@@ -74,10 +77,53 @@ export const InternalField = ({
...
         </Text>
       )}
+
+      {/* When completed is false, whitespace is rendered */}
+      {completed !== undefined && (
+        <AnimationWrapper
+          displayStatus={completed}
+          isError={!!(error && errorMsg)}
+        >
+          <StatusWrapper mt={'8px'}>
```
matdiak, line 87: I tried to use flexProps on Box but it doesn't actually allow a gap of 4px, so I kept it as StatusWrapper. 
Unless we are happy to add '4px' to `type Spacing`?

fleonard (approved):

PR #2809, opened 2023-06-22
Rebrand Buttons component

ellajay:
## Screenshot / video

image

## What does this do?

Updates the Buttons component and stories in line with the rebrand

- Update Button colours
- Deprecate `tertiary` style and replace with `fallback`
- Add `textBtn` style with optional `trailingIcon`

FE RFC [here](https://www.notion.so/getmarshmallow/Buttons-0f02e12343c049f2938819262b690fe9)
Design [here](https://www.figma.com/file/FqSaizGOyOzXFZNmZ4X0LGMn/branch/QtMe0nsVPpPkgqtdqd2Op2/%F0%9F%8D%AC-S'mores?type=design&node-id=13006-107162&mode=design&t=f0OWcvekbOrxyb72-0)

### DIFF
--- a/src/Button/Button.stories.tsx
+++ b/src/Button/Button.stories.tsx
@@ -21,7 +21,7 @@ export const Playground = Template.bind({})
 Playground.args = {
   primary: true,
   secondary: false,
-  tertiary: false,
+  fallback: false,
   disabled: false,
   loading: false,
   icon: '',
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -1,11 +1,9 @@
 import React, { FC, ReactNode, ButtonHTMLAttributes, forwardRef } from 'react'
 import styled, { css } from 'styled-components'
-import { darken } from 'polished'
 
-import { Color, theme } from '../theme'
+import { theme } from '../theme'
 import { focusOutline } from '../utils/focusOutline'
 import { useDeprecatedWarning } from '../utils/deprecated'
-import { LegacyButton } from './LegacyButton'
 import { Box } from '../Box'
 import { Loader } from '../Loader'
 import { Icon as IconComponent } from '../Icon'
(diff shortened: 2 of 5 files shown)


DeadEnglish (approved):

src/Button/Button.stories.tsx
```diff
@@ -21,7 +21,7 @@ export const Playground = Template.bind({})
 Playground.args = {
   primary: true,
   secondary: false,
-  tertiary: false,
+  fallback: false,
```
jacobmarshmallow, line 24: What does this fallback mean?

src/Button/Button.tsx
```diff
@@ -42,64 +38,45 @@ export const Button: FC<ButtonProps> = forwardRef<
...
   useDeprecatedWarning({
-    enabled: isLegacyButton,
-    title: 'Legacy Button',
+    enabled: tertiary,
+    title: 'Tertiary prop usage',
     message:
-      "You're using the legacy Button component. Please use the new Button by providing one of the following props: 'primary', 'secondary', 'tertiary'.",
+      "You're using the tertiary prop which is now deprecated. Please use the new 'fallback' prop instead, or 'primary' or 'secondary' as appropriate.",
     componentProps: props,
   })
```
DeadEnglish, line 63: If `tertiary` and `fallback` are similar or `fallback` is replacing it, we could just remove `tertiary` entirely.

jacobmarshmallow (approved):

src/Button/Button.stories.tsx
ellajay: It's a new style that we are adding to several components that will usually have a white/light coloured background. It is only intended to be used on colourful backgrounds where the primary and secondary styles would not provide enough contrast

DeadEnglish (approved):

PR #2836, opened 2023-06-28
Rebranded pagination

rowanhen:
## Screenshot / video

Screenshot 2023-06-28 at 17 26 35

## What does this do?

- update colours to be inline with designs
- adds `bgColor` on inactive state
- updates margin to always be `4px`


src/Pagination/Pagination.tsx
```diff
@@ -167,10 +192,28 @@ const Container = styled(Box)`
   }
 `
 
-const PageBox = styled(Box)<IPageBox>`
-  background: ${({ active }) => (active ? theme.colors.coconut : 'none')};
+const PageBox = styled.button<IPageBox>`
```
DeadEnglish, line 195: Probably worth calling this `PageButton` as it's no longer a `Box` 😄

DeadEnglish (approved):

PR #2840, opened 2023-06-29
Rebranded datepicker

rowanhen:
## Screenshot / video
Screenshot 2023-06-29 at 10 45 04

## What does this do?

- border radius to `16px`
- removal of outside border
- converts dates to buttons & adds onKeyPress events for accessibility


src/Datepicker/DatesList.tsx
```diff
@@ -78,11 +84,11 @@ interface IListItem {
   disabled?: boolean
 }
 
-const ListItem = styled.li<IListItem>`
+const ListItem = styled.button<IListItem>`
```
DeadEnglish, line 87: this should probably be `ListButton` and hte `container` can simply be a `div` rather than a `ul` as I don't think `button` as direct children of a `ul` semantically correct 😄

DeadEnglish (changes requested):

src/Datepicker/DatesList.tsx
rowanhen: 👀 👀 👀 , just pushed this change

DeadEnglish (approved):

PR #2950, opened 2023-08-03
Adding icon visual to radiogroup

rowanhen:
## Screenshot / video

Screenshot 2023-08-03 at 14 36 20

## What does this do?

- allows an to be passed into the rendered in the RadioGroup component


ellajay:
Can we call it `Icon` instead of `IconVisual`? Just as we already have a visual prop and to fit the naming convention used in other components. Also I'm not sure of the use case, but will we need the option to control the alignment of the icon?

ellajay:
@rowanhen I think it's fine to update this component to include an optional Icon, was just curious whether we'd be seeing more of these throughout the products :)

ellajay (dismissed):

DeadEnglish (approved):

PR #3106, opened 2023-09-21
fix: adding hideOverflow, rowActions bgColor, columnPadding, props to the Table

rowanhen:
## Screenshot / video

Before:
Screenshot 2023-09-21 at 09 05 35

After:
Screenshot 2023-09-21 at 09 08 34

Screenshot 2023-09-21 at 16 15 42

## What does this do?

- adds option to add `overflow: hidden` prop easily in the columns object, allowing for tooltips to show correctly.

### DIFF
--- a/src/Table/components/commonComponents.tsx
+++ b/src/Table/components/commonComponents.tsx
@@ -52,11 +52,16 @@ export const StyledHeaderCell = styled.th<TableStylesProps>`
 
 export const StyledCell = styled.td<TableStylesProps>`
   vertical-align: middle;
-  overflow: hidden;
   padding-left: 8px;
   padding-right: 8px;
   padding-top: 8px;
   padding-bottom: 8px;
+
+  ${({ hideOverflow }) =>
+    hideOverflow &&
+    css`
+      overflow: hidden;
+    `};
 
   ${({ noWrapContent }) =>
     noWrapContent &&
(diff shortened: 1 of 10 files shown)


src/Table/components/commonComponents.tsx
```diff
@@ -67,6 +72,7 @@ export const StyledCell = styled.td<TableStylesProps>`
   ${({ truncateContent }) =>
     truncateContent &&
     css`
+      overflow: hidden;
```
liampuk, line 75: Do we want this on truncateContent? I feel like that might be a usecase for tooltips 👀

rowanhen: yeah you're right, i'll try find a diff solution - the issue is with this at the moment, for truncateContent is we need the `overflow: hidden` to make use of the `text-overflow: ellipsis` property

if theres nothing super easy, it might make sense to just remove this option & have this logic in the actual column cell as we can customise as much as we want inside there

rowanhen: `overflow: clipped` solves it

rowanhen: we're so back

rowanhen: its over, i got excited and it didn't work - so i think we may aswell keep the prop - but for a tooltip or something like that, we should just handle that logic in the cell

PR #3182, opened 2023-10-24
Fix button loading disabled state

matdiak:
In rebrand work (#2809), we changed that the button is no longer disabled when loading by defualt. We discussed it in the FE meeting that it should not be the case.

This PR makes Button disabled when `loading` is `true` by default.


CHANGELOG.md
```diff
@@ -722,6 +726,7 @@
 ### Changed
 - Updated gap and styles on Row component
 
+[3.1.12]: https://github.com/marshmallow-insurance/smores-react/compare/v3.1.12...v3.1.13
```
rowanhen, line 729:
```suggestion
[3.1.13]: https://github.com/marshmallow-insurance/smores-react/compare/v3.1.12...v3.1.13
```

rowanhen (approved):

PR #3239, opened 2023-11-20
feat: adding semantic releases

rowanhen:
## Screenshot / video

N/A

## What does this do?

- adds semantic releases to `smores`
- tested NPM publishing locally using https://verdaccio.org/ (no idea if this will work when actually running here so its kinda just a 🤞)
- updates README
- adds preview-release script


DeadEnglish:
Nice work!

As part of this PR, could we do a manual release to see how our `changelog.md` is affected.

**Potential Improvement**
- using something like [commitlint](https://github.com/conventional-changelog/commitlint) to enforce conventional commits (I would love this to be everywhere)
-

DeadEnglish (approved):

PR #3282, opened 2023-11-29
fix: add in release-notes-generator

rowanhen:
## Screenshot / video

N/A

## What does this do?

- adds in the `release-notes-generator`
- and conventional commits

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "babel-eslint": "^10.1.0",
     "babel-jest": "^29.0.2",
     "babel-loader": "^9.1.0",
+    "conventional-changelog-conventionalcommits": "^7.0.2",
     "eslint": "^7.9.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "babel-eslint": "^10.1.0",
     "babel-jest": "^29.0.2",
     "babel-loader": "^9.1.0",
+    "conventional-changelog-conventionalcommits": "^7.0.2",
```
DeadEnglish, line 63: Do we need this package?

rowanhen: i think we do - im reading this [here](https://github.com/semantic-release/release-notes-generator#configuration:~:text=then%20subject)-,Configuration,-Options) and its saying its needed for that

DeadEnglish (approved):

PR #3326, opened 2023-12-11
breaking change: refactor Icon component

rowanhen:
## Loom demo
This is a demo showing the package is safe to use locally (done with yalc) - https://www.loom.com/share/f1637f327fd14a0d9d42810db9da3c4a?sid=cc04ec36-0a88-465b-a795-a9568641ba70

## Implementation plan & Breaking Changes

We have introduced the `Icons` to be passed to the Icons component.

As we have to use a valid Icon for anything to render. This should not introduce any breaking changes on the Smores component level. This _should_ only impact any snowflake components in our repositories. (An example can be seem in the Breadcrumbs component in AP).

in the below photo, A is the current version of S'mores, and B is after this refactor.

## Screenshot / video

you can see we have gained two icons that we were always available, but due to Icons being created automatically off the mappings list, we can actually see them now.

Screenshot 2023-12-11 at 14 27 14

## What does this do?

- implements changes from this doc: https://www.notion.so/getmarshmallow/Refactor-Icon-Implementation-3ab6ed279f154e0ba8caa5ec49771531
- allows us to use SVGs as react components now in our storybook build without any issues
- allows us to select icons when typing render
- adjusts types components using `<Icons />` to use the new `Icons` type
- does not change any existing snapshot tests 
- uses `vite-plugin-svgr` to handle conversion of svg files to react components at build time
- there is a following PR to actually bundle these files with vite & also move to vitest. (found here)

#### Related PRs - 
- https://github.com/marshmallow-insurance/smores-react/pull/3387

## Browsers tested on
I could not find any strict benchmark compatibilities to meet. So i first went as low as I could without these changes. I then replicated .
These were tested with yalc, using localhost & browserstack. I think browserstack is not ideal with testing on iPhone. I could not get any iPhone environments running. With or without these changes.

- Chrome: 106+
- Opera: 70+
- Edge: 80+
- Mozilla: 100+
- Android - Galaxy S8+
- Android - Pixel 2+


src/Icon/iconsList.ts
```diff
@@ -0,0 +1,415 @@
...
+  twitter: Twitter,
+  upload: Upload,
+  vandalism: Vandalism,
+  village: Village,
+  visa: Visa,
+  warning: Warning,
+  watch: Watch,
+  wave: Wave,
+  wellbeing: Wellbeing,
+  windscreen: Windscreen,
```
DeadEnglish, line 414: I would probably say for consistency we should have these in quotes even though it's not strictly needed.

src/Icon/iconsTypes.ts
DeadEnglish, line 1: we can remove this entire file in favour of doing the following with `iconsList.ts`:

`export type Icons = keyof typeof iconList`

This would get all the keys of the object and create a type from them, also allows for extensibility and ease of use by simply adding the icon to the list and it'll work!

I can see you're referencing `Icons` within that type so will need some fixing and palying with types. I would probably say we should allow the type to be inferred from the object by removing the `Record<Icons, FC<{ color: Color }>>` type.

DeadEnglish (changes requested):

src/Icon/assets/AccountBreakdown.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FC } from 'react'
+
+export const AccountBreakdown: FC = () => {
```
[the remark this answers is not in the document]
rowanhen: !!!!! i cant believe i didnt find this / think of this, i did svg { path: ${color} } failed and didnt think to go much further lmao

this fixes it all - will adjust and make these changes now

DeadEnglish (approved):

PR #3374, opened 2024-01-04
fix: validateDomNesting warnings in SupportMessage & Tooltip

rowanhen:
## Screenshot / video

N/A

## What does this do?

- fixes `validateDOMNesting(...): <div> cannot appear as a descendant of <p>` that would appear when using a ReactElement with SupportMessage description prop by adding in a ReactElement check
- fixes same error in Tooltip but instead by converting the UnderlinedText, <p /> component, to instead be a UnderlinedChild, <div /> component!
- adds labelBottom to LabelledText prop - to remove need for MarginlessLabelledText snowflake components

### DIFF
--- a/src/SupportMessage/SupportMessage.stories.tsx
+++ b/src/SupportMessage/SupportMessage.stories.tsx
@@ -28,7 +28,7 @@ WithCustomDescription.args = {
   type: 'info',
   title: 'A SupportMessage using the Link component',
   description: (
-    <Box>
+    <div>
       Some text rendered using a <Link href={''}>Link</Link>
     </Box>
   ),
--- a/src/SupportMessage/SupportMessage.stories.tsx
+++ b/src/SupportMessage/SupportMessage.stories.tsx
@@ -1,5 +1,4 @@
 import React from 'react'
-import { Box } from '../Box'
 import { Link } from '../Link'
 import { CollectionPage } from './Collection'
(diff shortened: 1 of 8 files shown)


src/SupportMessage/SupportMessage.stories.tsx
```diff
@@ -28,9 +29,9 @@ WithCustomDescription.args = {
   type: 'info',
   title: 'A SupportMessage using the Link component',
   description: (
-    <>
+    <Box>
```
DeadEnglish, line 32: there doesn't seem to be any benefit of importing `Box` so it might be worth just using a `div` 😄

DeadEnglish (approved):

src/SupportMessage/SupportMessage.stories.tsx
```diff
@@ -28,9 +29,9 @@ WithCustomDescription.args = {
...
       Some text rendered using a <Link href={''}>Link</Link>
-    </>
+    </Box>
```
rowanhen, line 34:
```suggestion
    </div>
```

src/SupportMessage/SupportMessage.stories.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
-import { SupportMessage, SupportMessageProps } from './SupportMessage'
+import { Box } from '../Box'
```
rowanhen, line 2:
```suggestion
```

PR #3390, opened 2024-01-11
breaking change: upgrade styled-components

rowanhen:
## Screenshot / video

N/A

## What does this do?

- bumps styled-components
- bumps `@types/react` to `18.2.0`
- bumps `@types/react-dom` to `18.2.0`


matdiak (commented):
For the styled components, can you include the migration docs in the PR and did you look into this one specifically? 
"dropped automatic prop filtering; use transient props ($ prefix) for stuff you don't want to be passed to child component / HTML

PR #3413, opened 2024-01-18
feat(dropdown): allow placeholder to be used as empty value

DeadEnglish:
## Screenshot / video
Screenshot 2024-01-18 at 15 33 24
Screenshot 2024-01-18 at 15 33 29
Screenshot 2024-01-19 at 10 58 26
Screenshot 2024-01-19 at 10 57 19

## What does this do?

- Adds an optional prop, `showDefaultOption` , to the `Dropdown` component to render a default selected option of "Select an option"
- Adds another optional prop, `customDefaultOption` that allows the user to override the "Select an option" value 
- Adds placeholder and show default option values to the dropdown `Collection.tsx`

## Ticket
- [Github Ticket](https://github.com/marshmallow-insurance/growth/issues/638)
- [Slack thread from Design](https://eatmarshmallows.slack.com/archives/C02U91D49C6/p1705658871408059)

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -24,6 +24,7 @@ export type DropdownItem = {
 
 export interface Props extends CommonFieldProps {
   placeholder?: string
+  showPlaceholderAsValue?: boolean
   name?: string
   value?: string | null
   defaultValue?: string
@@ -51,6 +52,7 @@ export const Dropdown = forwardRef(function Dropdown(
   {
     id: idProp,
     placeholder,
+    showPlaceholderAsValue = false,
     name,
     value: valueProp,
     defaultValue,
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -117,12 +119,22 @@ export const Dropdown = forwardRef(function Dropdown(
         >
           {hasOptGroups ? (
             <optgroup label={placeholder}>
-              <option value="" hidden>
+              <option
+                value=""
```
matdiak, line 123: Suggestion: can this just be hidden={!(placeholder && showPlaceholderAsValue)}

matdiak (commented):
My understanding was this this change will take effect on all inputs and we will impose this behaviour on all existing inputs?

Is it that this is potentially breaking what's stopping us from doing that?

src/Dropdown/Dropdown.tsx
DeadEnglish: The actual property is just `hidden` so any existence of it will trigger it to be hidden. adding `true` or `false` makes no difference so we need to when we don't want the `hidden` attribute, we need to entirely remove the prop.

DeadEnglish:
@matdiak The change for adding this to all inputs was for signup only, This shouldn't be a breaking change within S'mores as `showPlaceholderAsValue` is defaulted to false so will not change the behaviour of any other input in other applications.

matdiak (approved):

PR #3431, opened 2024-01-25
feat: add icons used in CMS

urbnzs:
## Screenshot / video

## What does this do?

I've added some icons that we use in CMS, in preparation for the S'mores update. 

To do in CMS:
- rename icon `maximiseWindow` to `maximise-window`, to match naming pattern

### DIFF
--- a/src/Icon/Icon.tsx
+++ b/src/Icon/Icon.tsx
@@ -1093,6 +1093,23 @@ export const Icon: FC<IconProps> = ({
         />
       </svg>
     )}
+    {render === 'clipboard' && (
+      <svg
+        xmlns="http://www.w3.org/2000/svg"
+        width="16"
+        height="16"
+        fill="none"
+      >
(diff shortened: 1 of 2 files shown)


src/Icon/Icon.tsx
```diff
@@ -1093,6 +1093,23 @@ export const Icon: FC<IconProps> = ({
...
     )}
+    {render === 'clipboard' && (
+      <svg
+        xmlns="http://www.w3.org/2000/svg"
+        width="16"
+        height="16"
+        fill="none"
+      >
+        <path
+          fill="#000"
```
rowanhen, line 1104: can we change any of these fill value to use the dynamic `theme.colors[color]` instead?

urbnzs: makes sense, changed them!

rowanhen (approved):

PR #3467, opened 2024-02-05
fix: fix NumberInput for currencies

liampuk:
## What does this do?

- Fixes how numbers are rounded for currencies, previously if you tried to enter 100.01, when typing 100.0 javascript would stringify this to 100
- `strict` doesn't seem to be used anywhere so I've removed it, though keeping normalising with min & max

### DIFF
--- a/src/NumberInput/NumberInput.tsx
+++ b/src/NumberInput/NumberInput.tsx
@@ -27,7 +27,6 @@ export interface Props extends CommonFieldProps {
   onBlur?: (e: FocusEvent<HTMLInputElement>) => void
   min?: number
   max?: number
-  strict?: boolean
   roundCurrency?: boolean
   step?: number
 }
@@ -60,7 +59,6 @@ export const NumberInput = forwardRef(function NumberInput(
     roundCurrency,
     min = -999999,
     max = 999999,
-    strict,
     step = 0,
     disabled = false,
     error = false,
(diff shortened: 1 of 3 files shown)


src/NumberInput/NumberInput.tsx
```diff
@@ -91,28 +89,21 @@ export const NumberInput = forwardRef(function NumberInput(
...
   }
 
-  const handleStrictValue = (event: number): number => {
-    if (isInRange(event)) {
+  const formatCurrency = (event: string) => {
+    const decimalIndex = event.indexOf('.')
+    if (decimalIndex >= 0 && event.length > decimalIndex + 1) {
+      const fractionalString = event.substring(decimalIndex + 1).substring(0, 2)
+      return `${event.substring(0, decimalIndex)}.${fractionalString}`
+    } else {
```
rowanhen, line 97: think using a number here for the event may make this formatting here easier to do - some comments here would also be helpful too

liampuk: The issue we had was that using number caused some weird behaviours, typing `20.0` would be stringified to `20`, and the currency formatting didn't prevent adding extra zeros (`20.1000000`) because as a number it's just `20.1`, so the setState/whatever function passed into the component wouldn't run. Using string fixes a lot of these issues but I know it's weird in a number input component

Maybe something to chat about in FE guild tomorrow ⚔️

DeadEnglish (commented):
One QoL change we can make is adding `inputmode="numeric"` to the input component. This makes no difference on desktop but will show the number input on mobile for users 😄

src/NumberInput/NumberInput.tsx
liampuk: Could we maybe do that in a separate pr? This function is just to handle currencies really

rowanhen (approved):

PR #3484, opened 2024-02-09
breaking change: deprecate props in NumberInput and add CurrencyInput…

liampuk:
… component

## Screenshot / video

image

## What does this do?

- Deprecate strict and roundCurrency props in NumberInput
- Add a new CurrencyInput component. This component is similar to NumberInput but clips decimal places at 2 and represents the value with a string so no decimal places are lost


src/CurrencyInput/CurrencyInput.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  const applyMinMax = (event: string) => {
+    const value = Number(event)
+    if(min && value < min) return min.toString()
+    if(max && value > max) return max.toString()
+    return event
+  }
+
+  const handleChange = (event: string) => {
```
urbnzs, line 81: I think it's a bit confusing that the param is is called `event`, when in fact you're passing the `e.currentTarget.value` to this function which is a string, so I'd rename it to `value` or something like that instead. Same goes to the `formatCurrency` and the `applyMinMax` functions.

src/CurrencyInput/CurrencyInput.tsx
```diff
@@ -0,0 +1,123 @@
...
+          onChange={(e: FormEvent<HTMLInputElement>) => {
+            onChange && handleChange(e.currentTarget.value)
+            onInputChange && onInputChange(e)
+          }}
+          onBlur={onBlur}
+        />
+      </Box>
+    </Field>
+  )
+})
```
urbnzs, line 123: missing newline 🚓

urbnzs (approved):

PR #3520, opened 2024-02-26
feat: add rich text viewer

liampuk:
## Screenshot / video

image

### 🖼️ Adds Rich Text editing  to Smores 🖼️ 

- Uses [DOMPurify](https://github.com/cure53/DOMPurify) as recommended by the [MDN docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Safely_inserting_external_content_into_a_page#working_with_html_content) to sanitise html
- Resets styles to support plain html tags, or inline styles from emails


package.json
```diff
@@ -82,9 +82,11 @@
     "vitest": "^1.2.2"
   },
   "dependencies": {
+    "@types/dompurify": "^3.0.5",
```
DeadEnglish, line 85: these can be moved to the `devDependencies` 😄

src/RichText/__tests__/RichText.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const { container } = render(
+    <RichText htmlString="<h1>header</h1><p>paragraph</p>" />,
+  )
+  expect(container.innerHTML).toContain('header')
+  expect(container.innerHTML).toContain('paragraph')
+})
+
+test('strips script tag', () => {
+  const { container } = render(
+    <RichText htmlString="<script>angry javascript</script><h1>header</h1><p>paragraph</p>" />,
```
DeadEnglish, line 15: 😠

DeadEnglish (approved):

PR #3536, opened 2024-03-04
skip lib check

liampuk:
## What does this do?

- skip type checking of node modules
- Looks like facebook use less strict settings for their typescript compiler, ours fails on `@lexical/react` when run locally
- [issue discussion](https://github.com/facebook/lexical/issues/2528)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -28,7 +28,8 @@
     "strictNullChecks": true,
     "noUnusedLocals": true,
     "allowSyntheticDefaultImports": true,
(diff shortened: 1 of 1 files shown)


rowanhen (approved):

jacobmarshmallow:
Does it still check our calls to the libs are valid?

liampuk:
@jacobmarshmallow Yes it will do, it skips type declaration files (`d.ts`), should only be in `node_modules`

PR #3555, opened 2024-03-11
breaking change: styled components v6

rowanhen:
## Screenshot / video

N/A

## What does this do?

- converts use of `as` to `forwardedAs`
- bumps styled-components to v6
- renames `bgColour` prop to `bgColor`
- closes: https://github.com/marshmallow-insurance/smores-react/pull/3382

## Migration info - [v6 styled-component migration info](https://styled-components.com/releases#v6.0.0)

-> audit use of `as="XXX"` and address if this needs to be moved to `forwardedAs="XXX"` instead
-> remove `@types/styled-components` package from our FE repos
-> with the release of this, we should introduce the below code to our repos

```
import isPropValid from '@emotion/is-prop-valid';
import { StyleSheetManager } from 'styled-components';


function MyApp() {
    return (
        <StyleSheetManager shouldForwardProp={shouldForwardProp}>
            {/* other providers or your application's JSX */}
        </StyleSheetManager>
    )
}


// This implements the default behavior from styled-components v5
function shouldForwardProp(propName, target) {
    if (typeof target === "string") {
        // For HTML elements, forward the prop if it is a valid HTML attribute
        return isPropValid(propName);
    }
    // For other elements, forward all props
    return true;
}
```

### Pre-release package

-  "@mrshmllw/smores-react": "^7.0.0-crumbs-sc-v6.1"
- [7.0.0-crumbs-sc-v6.1](https://www.npmjs.com/package/@mrshmllw/smores-react/v/7.0.0-crumbs-sc-v6.1)

####
#### Pre-migration discussion

- [x] discuss whether we should migrate to transientProps or use this recommendation, to ease the migration https://arc.net/l/quote/qbwkeecm

-> outcome: carry out the transient prop changes only in a secondary PR & have those be breaking changes


src/BrandCard/BrandCard.tsx
```diff
@@ -18,7 +18,7 @@ export type BrandCardProps = {
   visualHeight?: string
   /** render image at bottom of card */
   visualBottom?: boolean
-  bgColour: Color
+  bgColor: Color
```
urbnzs, line 21: have you checked if any of our apps use this component with the `bgColour` prop? I agree that it should be renamed, but it might be worth mentioning in the changelog to avoid ui issues.

rowanhen: :eyes: i had a look and its not used at all in any of our repos - ill add a comment to the PR mentioning this tho so semantic-releases will add it to the changelog

urbnzs (approved):
I tried the pre-release package on CMS, everything seems to be working 🚀

PR #3558, opened 2024-03-11
fix: handle 3 digits on pagination component

liampuk:
## Screenshot / video

Screenshot 2024-03-11 at 11 05 32

## What does this do?

- handle 3 digits on pagination component

### DIFF
--- a/src/Pagination/Pagination.tsx
+++ b/src/Pagination/Pagination.tsx
@@ -193,17 +193,16 @@ const Container = styled(Box)`
 `
 
 const PageBox = styled.button<IPageBox>`
-  background: ${({ active }) =>
-    active ? theme.colors.custard : theme.colors.cream};
-    cursor: pointer;
-    display: flex;
-    -webkit-box-pack: center;
-    justify-content: center;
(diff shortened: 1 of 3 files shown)


src/Pagination/Pagination.tsx
```diff
@@ -193,14 +193,16 @@ const Container = styled(Box)`
 `
 
 const PageBox = styled.button<IPageBox>`
-  background: ${({ active }) =>
-    active ? theme.colors.custard : theme.colors.cream};
+  background: ${({ active }) => active ? theme.colors.custard : theme.colors.cream};
   cursor: pointer;
   display: flex;
+  -webkit-box-pack: center;
```
david-gomes5, line 199: `box-pack` looks to be deprecated and not recommended. If I understand correctly, you might be trying to vertically centre the element? 

Would `place-items`, `place-content` or `place-self` achieve the same aims?

liampuk: ah thanks for spotting, I copied these from dev-tools and didn't notice this was included, we don't need this field

PR #3622, opened 2024-03-25
chore: add semantic-release dry run on main merge

rowanhen:
## Screenshot / video

N/A

## What does this do?

- adds `npx run semantic-release --dry-run` to the on merge command
- this is to hopefully catch any issues with publishing before running the GHA workflow (also saves use using the `preview bump & publish` GHA everytime if we dont want to)

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -19,6 +19,11 @@ jobs:
         run: |
           npm ci
           npm run build-storybook
+      - name: Preview Publish 🦫
+        env:
+          GITHUB_TOKEN: ${{ secrets.MARSHMALLOW_CI_PAT }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -19,6 +19,11 @@ jobs:
         run: |
           npm ci
           npm run build-storybook
+      - name: Preview Publish 🦫
+        env:
+          GITHUB_TOKEN: ${{ secrets.MARSHMALLOW_CI_PAT }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx semantic-release --dry-run
```
david-gomes5, line 26: Just to understand the value for this - if there is an issue that's found in the dry run, do we know if the workflow would quit on this step or carry on?

david-gomes5 (approved):
Have a query but it's irrelevant to approving PR.

PR #3654, opened 2024-04-09
breaking change: allow update default value for rich text editor

liampuk:
## What does this do?

- Change `defaultValue` to `value` in rich text editor (breaking change)
- Allow the value provided to the rich text editor to be updated - for use in email templating


snegyeliczky (approved):
Perfetto 💯

PR #3687, opened 2024-04-24
feat: add padding props types to `Icon`

david-gomes5:
## What does this do?

- Add the typing for padding props to allow downstream to use padding props in `Icon` without typescript errors

`Container` already supports taking in padding we just don't type it.
`Container` is built from `Box`. `Box` implements padding styles from padding props


david-gomes5:
Closing as realising that exposing these types might be bad practice for styling our icons

PR #3706, opened 2024-05-03
Feat: export useBodyScrollLock hook

matdiak:
## What does this do?
Exports useBodyScrollLock so it can be used outside of smores


DeadEnglish:
Now that we're using this as an export, could we move this to the `hooks` folder rather than nested inside the `Modal` component.

matdiak:
@DeadEnglish moved it to utils with other hooks unless you think creating a hook dir?

DeadEnglish:
There should already be a [hook directory](https://github.com/marshmallow-insurance/smores-react/tree/main/src/hooks) in the `src` folder 😄

matdiak:
@DeadEnglish oh ye, somethow I didn't see it

DeadEnglish (approved):

PR #3707, opened 2024-05-03
Feat: Add caret-up icon

matdiak:
## What does this do?
Adds a caret-up icon 
caret-up


DeadEnglish:
> can we instead use the rotate prop on the icon component instead of this?

Ideally this unless the icon is different 😄

matdiak:
The Icon component is used internally within the button. Do you still prefer to pass a prop to `Button` to rotate?

DeadEnglish:
I think we can just style this with a styled component and rotate in css via the `svg`

DeadEnglish (approved):

PR #3736, opened 2024-05-13
feat(search-input): implement fuzzy searching capability

DeadEnglish:
## Screenshot / video

[loom demonstration](https://www.loom.com/share/0015aef22ac342348627ed40247a83f1)

## What does this do?

Updating the search results to now use fuzzy searching via `fuse.js` if a prop is enabled.

- Adds an optional `useFuzzySearch` prop to enable fuzzy search
- Adds an optional `onNotFound` prop that runs every time there are no results
- Adds an optional `notFoundComponent` prop to pass a component to render when no results are shown


src/SearchInput/SearchInput.tsx
```diff
@@ -58,18 +66,31 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
...
-
     const [searchQuery, setSearchQuery] = useState<string | null>(null)
 
+    const fuse = useMemo(() => {
+      const searchKeys = ['label', 'value']
+      return new Fuse(searchList, {
+        shouldSort: true,
+        keys: searchKeys,
+        findAllMatches: true,
+        minMatchCharLength: 2,
```
DeadEnglish, line 77: I've set some defaults here. as we only have a `label` and `value` on the options, these are the only keys we can use so these can remain static. The rest can potentially refined at a later date if we need to but in tl;dr:

**minMatchCharLength**: Requires at least 2 characters to find a match
**should sort**: auto sorts results based on the "weight" given to them
**findAllMatches**: finds all matches even if an exact match is found e.g. searching for if a list contains both `Analyst` and `Data Analyst` and the user searches for `Analyst`, both will be returned but `Analyst` will be top as it is an exact match

src/SearchInput/SearchInput.tsx
```diff
@@ -23,11 +25,14 @@ export interface SearchInputProps extends CommonFieldProps {
   placeholder?: string
   searchList: SearchInputItem[]
   onFound: (element: string) => void
+  onNotFound?: () => void
+  notFoundComponent?: ReactNode
```
david-gomes5, line 29: Would we be able to add JSDocs to these new props? Adding JSDocs above the props will surface the doc on tooltips and should also display on storybook.
e.g.
```suggestion
  /**
  * Callback that runs once when search finds zero results
  */
  onNotFound?: () => void
  notFoundComponent?: ReactNode
```

src/SearchInput/components/NoOptions.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FC, ReactNode, useEffect } from 'react'
+
+interface Props {
+  onNotFound?: () => void
+  notFoundComponent?: ReactNode
+}
+
+export const EmptyResults: FC<Props> = ({ onNotFound, notFoundComponent }) => {
+  useEffect(() => {
+    onNotFound && onNotFound()
```
david-gomes5, line 10: I'm not sure if the new ESLint rule would error on this but you can use optional chaining here

```suggestion
    onNotFound.?()
```

david-gomes5 (commented):
Great stuff - I don't really have time atm to give it a good review, I just have some quick suggestions.

Feel free to skip my review if someone else approves, otherwise will revisit for a review when I get some time

src/SearchInput/SearchInput.tsx
DeadEnglish: I've added the JSDocs but they're not rendering and I have a feeling it's to do with the use of `forwardRef` for our form components. Somewhat related, I had to set the all of the args within storybook as they weren't properly being created due to [this unresolved bug](https://github.com/storybookjs/storybook/issues/21686)

src/SearchInput/SearchInput.tsx
```diff
@@ -11,23 +12,39 @@ import { CommonFieldProps } from '../fields/commonFieldTypes'
...
+  /**  Optional callback to run on blur */
   onBlur?: (e: FocusEvent<HTMLInputElement>) => void
+  /**  Optional default value for input */
   value?: string
+  /**  Optional boolean to move results to a realtive position */
   resultsRelativePosition?: boolean
+  /**  optional boolean to add border to results */
   resultsBorder?: boolean
+  /** optional boolean to enable fuzzy search via fuse.js */
+  useFuzzySearch?: boolean
```
matdiak, line 47: Suggestion: maybe calling it `enableFuzzySearch` would be better as it avoid using the `use` keyword reserved for hooks?

src/SearchInput/SearchInput.tsx
```diff
@@ -58,18 +78,31 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
       initialState: null,
       stateProp: value,
     })
-
     const [searchQuery, setSearchQuery] = useState<string | null>(null)
 
+    const fuse = useMemo(() => {
+      const searchKeys = ['label', 'value']
+      return new Fuse(searchList, {
+        shouldSort: true,
```
matdiak, line 86: Nit: this is set to true by default unless you want to be explicit? [here](https://www.fusejs.io/api/options.html)

src/SearchInput/storybook/searchList.ts
```diff
@@ -0,0 +1,30 @@
+export const searchList = [
```
matdiak, line 1: Nit: Maybe we could make that list more elaborate so we can see how the fuzzy search actually works?

src/SearchInput/SearchInput.tsx
```diff
@@ -58,18 +78,31 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
       initialState: null,
       stateProp: value,
     })
-
     const [searchQuery, setSearchQuery] = useState<string | null>(null)
 
+    const fuse = useMemo(() => {
```
matdiak, line 83: Question: have you considered adding the search function as a prop so that it is treeshakeable like @david-gomes5 suggested?

DeadEnglish: I did, but after the discussion with design during the reviews this will most likely be rolled out as a default and move to more of an "opt out" implementation, so I'm not overly concerned with it being tree shakeable

src/SearchInput/storybook/searchList.ts
DeadEnglish: good idea, I've implemented fruits, enjoy! 🍓

matdiak (approved):

PR #3744, opened 2024-05-15
fix(search-input): propagate search term usage to not found function

DeadEnglish:
## Screenshot / video

https://github.com/marshmallow-insurance/smores-react/assets/5758372/a891aab4-dad2-42b6-80ca-7fbc26d32b25

## What does this do?

Fixing an issue where we could not use the search term entered into the search input component as this is a local state variable.


src/SearchInput/SearchInput.tsx
```diff
@@ -30,7 +30,7 @@ export interface SearchInputProps extends CommonFieldProps {
   /**  callback to handle found item click */
   onFound: (element: string) => void
```
david-gomes5, line 31: I know `onFound` was probably created before this work started but it seems like this callback is called when an option is selected?

I initially confused `onFound` with results being found, not selected. I think this prop _should_ be changed to something like `onSelectedOption` or `onSelectChange`. Although, probably not something for this PR (as assuming will be breaking change)

david-gomes5 (approved):

src/SearchInput/components/NoOptions.tsx
david-gomes5, line 1: Can we rename this file to from `NoOptions` to `EmptyResults`?

david-gomes5 (changes requested):
`src/SearchInput/components/NoOptions.tsx`

Can we rename this file to from `NoOptions` to `EmptyResults`?

david-gomes5 (approved):
Just a small request on file name change, otherwise LGTM 👍

src/SearchInput/SearchInput.tsx
DeadEnglish: agreed, I think when we make the breaking change to default fuzzy search, we can make that breaking change as well 🫡

PR #3751, opened 2024-05-16
feat: adding fuzzySearchOptions prop

rowanhen:
## Screenshot / video

N/A

## What does this do?

- allows us to pass fuzzySearchOptions to the SearchInput so we can customise the how fuse.js matches and works inside project repos
- adds industries and occupations stories to test fuzzySearch on

### DIFF
--- a/src/SearchInput/SearchInput.tsx
+++ b/src/SearchInput/SearchInput.tsx
@@ -1,3 +1,4 @@
+import Fuse, { IFuseOptions } from 'fuse.js'
 import React, {
   ChangeEvent,
   FocusEvent,
@@ -13,11 +14,15 @@ import { Input, StyledFrontIcon } from '../fields/components/CommonInput'
 import { useUniqueId } from '../utils/id'
 import { useControllableState } from '../utils/useControlledState'
 import { SearchOptions } from './components/SearchOptions'
-import Fuse from 'fuse.js'
 
 export type SearchInputItem = {
   label: string
   value: string
+  // Wanted to refactor this component to allow for a generic here
+  // but would take far too long since this is using forwardRef which
+  // complicates things a fair bit
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  tags?: any[]
 }
 
 export interface SearchInputProps extends CommonFieldProps {
@@ -45,6 +50,7 @@ export interface SearchInputProps extends CommonFieldProps {
   resultsBorder?: boolean
   /** optional boolean to enable fuzzy search via fuse.js */
   enableFuzzySearch?: boolean
+  fuzzySearchOptions?: IFuseOptions<SearchInputItem>
 }
 
 export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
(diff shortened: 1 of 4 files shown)


src/SearchInput/SearchInput.tsx
```diff
@@ -81,27 +96,22 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
...
+      return new Fuse(searchList, fuzzySearchOptions)
     }, [searchList])
 
     const filteredList = useMemo(() => {
       if (searchQuery === null || searchQuery === '') {
         return searchList
       }
 
-      if (enableFuzzySearch) {
+      if (enableFuzzySearch || !!fuzzySearchOptions) {
```
rowanhen, line 107: added this here - so we're not relying on two props for this to work, unsure how people feel about this, but i did spend 20mins trying to work out why this wasn't working then realised i didnt have enableFuzzySearch as true

src/SearchInput/SearchInput.tsx
```diff
@@ -45,6 +50,8 @@ export interface SearchInputProps extends CommonFieldProps {
   resultsBorder?: boolean
   /** optional boolean to enable fuzzy search via fuse.js */
   enableFuzzySearch?: boolean
+  /** optional config of fuzzySearchOptions */
```
david-gomes5: Should this be:
```suggestion
  /** optional config of fuzzy search */
```

I don't think this can be a config of itself

src/SearchInput/SearchInput.tsx
DeadEnglish: Because `fuzzySearchOptions` is provided as a default in this component, this would always be true. Imo we would only need `enableFuzzySearch` as the variable here.

rowanhen, line 107:
```suggestion
      if (enableFuzzySearch) {
```

src/SearchInput/SearchInput.tsx
```diff
@@ -81,27 +96,22 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
...
-      if (enableFuzzySearch) {
+      if (enableFuzzySearch || !!fuzzySearchOptions) {
         return fuse.search(searchQuery).map(({ item }) => item)
       }
 
       return searchList.filter(({ label }) =>
         label.toLowerCase().includes(searchQuery.toLocaleLowerCase()),
       )
-    }, [searchQuery, enableFuzzySearch])
+    }, [searchQuery, enableFuzzySearch, !!fuzzySearchOptions])
```
rowanhen, line 114:
```suggestion
    }, [searchQuery, enableFuzzySearch])
```

src/SearchInput/SearchInput.tsx
rowanhen, line 53:
```suggestion
  /** optional config of fuzzySearchOptions, enableFuzzySearch is required for this to work */
```

src/SearchInput/SearchInput.tsx
```diff
@@ -45,6 +50,19 @@ export interface SearchInputProps extends CommonFieldProps {
   resultsBorder?: boolean
   /** optional boolean to enable fuzzy search via fuse.js */
   enableFuzzySearch?: boolean
+  /** optional config of fuzzy search
+   *  passing a value to this prop, automatically enabled fuzzy search
```
rowanhen, line 54:
```suggestion
   *  passing a value to this prop, automatically enables fuzzy search
```

PR #3755, opened 2024-05-20
feat(SearchInput): UI and accessibility improvements

rowanhen:
## Screenshot / video

https://github.com/marshmallow-insurance/smores-react/assets/57456071/621b9281-2dc3-4ba5-a591-44a9fbc53d3c

## What does this do?

- adds keyboard updown navigation
- adds ability to tab through the search results
- adds an optional clearButton prop
- adjusts the ui of the search results
- auto close the search results when clicking outside of the search input


src/SearchInput/SearchInput.tsx
```diff
@@ -150,59 +166,147 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
...
     const handleSelect = (nextValue: SearchInputItem): void => {
       updateSearchQuery(null)
-
       setSelectedValue(nextValue.label)
       onFound(nextValue.value)
     }
 
+    const handleClearSearch = () => {
+      updateSearchQuery(null)
+      setSelectedValue(null)
```
david-gomes5, line 195: Should we also be clearing the selected value if the user types further in the `SearchInput`?

I'm guessing we don't want a customer to submit a form with a selected value that is different to what's showing up in the `SearchInput`

I've recorded a weird interaction that happens when you manually clear the input after selecting a value:

https://github.com/marshmallow-insurance/smores-react/assets/63399235/b5a9bd29-1bb2-495a-ba57-1b9b524a3f20

rowanhen: yeah it’s a bit funky but I think this is the current behaviour right now - it does fix itself once you delete and then begin typing again

david-gomes5: Are we losing intended functionality if we clear `selectedValue` when a user starts typing again?

I'm opposed to use a component like this on AP with this current functionality. I wouldn't want agents to unknowingly submit values.

A similiar edgecase that I haven't highlighted in my recording is when we don't completely clear the input e.g. it could be 'Aba' - user goes to submit thinking it's going to submit 'Aba' but it submits 'Abattoir'

rowanhen: that seems to be the current behaviour of the search component right now

im open to a solution but think its worth doing as its own change - as im already introducing functionality here and dont want to alter the existing behaviour much more

rowanhen: thinking more about this, i think you're right and this should probably be done as this work - as we are now auto-closing the SearchOptions here, implying the value _could_ be okay

where as before, we would not close the SearchOptions until there was a valid value from the list inside

rowanhen: Will refactor this

david-gomes5 (approved):
Good stuff, Just 1 opinionated suggestion that you could ignore:
- I still think we should hide the 'x' button if there's nothing to clear

PR #3829, opened 2024-06-17
feat: Allow autocomplete and inputmode for TextInput

matdiak:
## Screenshot / video
[Demo](https://www.loom.com/share/0dd53f211b794fec952d8107d20dd4c6)

## What does this do?
- Allows TextInput to take autocomplete and inputmode attribute.
- autocomplete specifies what data we are expecting ([mdn](https://www.w3schools.com/tags/att_input_autocomplete.asp))
- input mode specifies which keyboard to show ([mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode))

Part of [M1 date of birth input](https://github.com/orgs/marshmallow-insurance/projects/7/views/29?pane=issue&itemId=67350316)

PR #3833, opened 2024-06-18
fix: replace reserved keywords with namespaces

matdiak:
## What does this do?
`autoComplete` and `inputMode` being react reserved words caused issues when using the smores as a library. It works in storybook but does not from sign-up. Hence renaming them.

Follows on 
- https://github.com/marshmallow-insurance/smores-react/pull/3829

### DIFF
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -34,8 +34,8 @@ type InputProps = (
       onInputChange: (e: FormEvent<HTMLInputElement>) => void
     }
 ) & {
-  autoComplete?: AutoComplete
-  inputMode?: InputMode
+  autoCompleteAttr?: AutoComplete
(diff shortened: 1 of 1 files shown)


DeadEnglish (approved):

david-gomes5:
@matdiak I'm curious to gain a further understanding of this. The html attribute version of inputMode is `inputmode`.

I'm trying to understand in what situation `inputMode` becomes a 'reserved word' - react docs recommend passing props that are for a attribute counter-part with a camel-casing naming convention like we do with `inputMode`.

The only props that are actually reserved in react and aren't passed through to the component should be `key` and `ref` ([old docs](https://legacy.reactjs.org/warnings/special-props.html)).

What environment does this issue happen in? are we maybe using styled components when this issue happens?

PR #3841, opened 2024-06-21
feat(TextDateInput): Add new precise date input component

ellajay:
## Screenshot / video

image

## What does this do?

Adds a new `TextDateInput` component that allows for a precise date to be entered in specific `day`, `month` and `year` input fields


src/TextDateInput/TextDateInput.tsx
```diff
@@ -0,0 +1,121 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { TextInput } from '../TextInput'
+import { Box } from '../Box'
+import { Fieldset, FieldsetProps } from '../fields/Fieldset'
+
+export type FullDateFormat = {
```
matdiak, line 8: I actually renamed in in sign-up. We could match is so that it is `DateObjectType`

src/TextDateInput/TextDateInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+  day: string | null
+  month: string | null
+  year: string | null
+}
+
+type ValidationErrorType =
+  | 'day'
+  | 'month'
+  | 'year'
+  | 'all'
```
matdiak, line 18: I think there will be more errors types encopassed under `all`. Maybe a better approach would be to pass an array as a prop for example, fieldsWithError: ['day', 'month'] and we will evaluate it whereever we use it?

src/TextDateInput/storybook/TextDateInput.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+} from '../TextDateInput'
+
+export default {
+  title: 'Text Date Input',
+  component: TextDateInput,
+  argTypes: { onChange: { action: 'changed' } },
+}
+
+const Template: FC = (props: TextDateInputProps) => {
+  const [_value, setValue] = useState<FullDateFormat>({
```
matdiak, line 16: Why is it `_value`?

ellajay: This was left over from a previous approach where there was already a `value` defined, we can use `value` now though so I'll update this

PR #3850, opened 2024-07-01
feat(Table-RowAction): Allow passing `title` to a row action

david-gomes5:
## What does this do?

Currently unable to add the `title` prop to a row action in a table, this change fixes this. Row action Icons ar currently defaulting to the `icon-button`.

- Allow adding `title` to row action


david-gomes5:
Buttons found in the invoice table for the policy payment screen on agent portal have a default title that cannot be changed.

This change should allow us to change the title for those buttons.

rowanhen (approved):

PR #3964, opened 2024-09-09
feat: add onAutoFill prop to TextInput

rowanhen:
## What does this do?

- adds onAutoFill prop to the TextInput component
- checks for onAutoFillStart animation and to identify when autofill occurs
- this is needed as we use onBlur events for the majority of our fields, however, this does not trigger when a user uses autofill as the browser does not focus the input

NOTE:
- there is one issue with this solution demonstrated in the video though sometimes we can emit an event if a user has a partially entered value, and hovers over a choice for 2s (time taken for the animation duration to end) - not sure this is a huge issue since we will fire again once the user selects the option anyways

## Screenshot / Video

https://github.com/user-attachments/assets/bf3591dd-74e2-40ed-83c4-e6a32edda431

### DIFF
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -1,4 +1,10 @@
-import React, { FocusEvent, FormEvent, ForwardedRef, forwardRef } from 'react'
+import React, {
+  FocusEvent,
+  FormEvent,
+  ForwardedRef,
+  forwardRef,
+  HTMLInputAutoCompleteAttribute,
+} from 'react'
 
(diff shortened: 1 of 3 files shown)


DeadEnglish (approved):

david-gomes5:
@rowanhen for future reference, can we add the caveat mentioned in the PR description around how the autocomplete event fires on hover as a JSDoc for the prop?

Consider a new developer without context assuming that the event only fires when an autocomplete action is completed e.g. selecting a autocomplete option

PR #4028, opened 2024-10-10
feat(SupportMessage): Add `ref` prop to `SupportMessage` component

david-gomes5:


david-gomes5:
No needed as there's a workaround to wrap SupportMessage in an element that can be captured instead.

PR #4036, opened 2024-10-16
fix(chip): icon size and min/max width

DeadEnglish:
## What does this do?
Fixes the icon size and min/max width for the Chip component

### Old
Screenshot 2024-10-16 at 09 22 25
Screenshot 2024-10-16 at 09 25 48

### New
Screenshot 2024-10-16 at 09 22 29
Screenshot 2024-10-16 at 09 23 41

- [Notion ticket](https://www.notion.so/getmarshmallow/Chip-a3d2a61ef2d2414189f6e8ed16d3edcd?pvs=4)
- [Figma design](https://www.figma.com/design/FqSaizGOyOzXFZNmZ4X0LGMn/%F0%9F%8D%AC-S'mores?m=auto&node-id=14411-158465)

## Testing
- Visual

### DIFF
--- a/src/Chip/Chip.tsx
+++ b/src/Chip/Chip.tsx
@@ -63,7 +63,7 @@ export const Chip: FC<ChipProps> = forwardRef<HTMLButtonElement, ChipProps>(
           {icon && (
             <IconComponent
               render={icon}
-              size={20}
+              size={16}
               color={primary ? 'liquorice' : 'cream'}
             />
           )}
@@ -90,7 +90,8 @@ const Container = styled(Box)<IButton>(
     justify-content: ${$icon ? 'space-evenly' : 'center'};
     line-height: 100%;
     padding: 8px 16px 8px ${$icon ? '8px' : '16px'};
-    width: 98px;
+    min-width: 98px;
+    max-width: 120px;
     min-height: 40px;
     cursor: ${disabled || $loading ? 'not-allowed' : 'pointer'};
     opacity: ${disabled ? '0.5' : '1'};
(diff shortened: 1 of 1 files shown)


ellajay:
@DeadEnglish Did you check all instances of Chip across our repos? In case the `max-width` causes an issue for any existing instances

DeadEnglish:
> @DeadEnglish Did you check all instances of Chip across our repos? In case the `max-width` causes an issue for any existing instances

Yep, SUF is our only usage of this component as far as I could see so shouldn't effect anything else 😅

ellajay (approved):

src/Chip/Chip.tsx
```diff
@@ -90,7 +90,8 @@ const Container = styled(Box)<IButton>(
     justify-content: ${$icon ? 'space-evenly' : 'center'};
     line-height: 100%;
     padding: 8px 16px 8px ${$icon ? '8px' : '16px'};
-    width: 98px;
+    min-width: 98px;
+    max-width: 120px;
```
david-gomes5, line 94: Is there a reason why we give it a `max-width`?

Does this not make this component not very flexible when used for chips with long text?

david-gomes5: Another point, I'm guessing `max-width` would be problematic for users that use big text (for accessibility)

DeadEnglish: good point tbf, by design these chips are supposed to be small with one word text such as "add"/"added" but and we shouldn't be using it for anything larger. That being said, I've removed the constraint incase we find ourselves in that position 😅 

Screenshot 2024-10-16 at 10 11 38

PR #4067, opened 2024-10-31
feat(SupportMessage): Add `rightSideComponent` prop

david-gomes5:
## What does this do?
- Adds `rightSideComponent` prop, allowing right side of the `SupportMessage`'s content to be configurable
- Updated component story, using new [Component Story Format (CSF)](https://storybook.js.org/docs/api/csf) API

Needed to add text on the right side of the component (see below for images and a link to designs)
## Screenshot / Video

https://github.com/user-attachments/assets/b7c8f922-ad23-4466-8385-259a0f323392

> Storybook demo

Screenshot 2024-10-31 at 15 08 15

- [Figma Designs](https://www.figma.com/design/SB6QnXvPF55xSNs3ohsK5W/Customer-Profile?m=auto&node-id=4942-89422&t=9mzKmwHzKdpJFDxM-1)
- closes https://github.com/marshmallow-insurance/multi-product/issues/662
- Used in https://github.com/marshmallow-insurance/uk-auto-agent-portal-www/pull/7339

### DIFF
--- a/src/SupportMessage/SupportMessage.tsx
+++ b/src/SupportMessage/SupportMessage.tsx
@@ -62,6 +62,10 @@ export type SupportMessageProps = {
   className?: string
   description: string | ReactElement
   onClick?: MouseEventHandler
+  /**
+   * Right side content, usually an icon or a button
+   */
+  rightSideComponent?: ReactElement
   type: SupportMessageType
   title?: string
 } & MarginProps
@@ -70,7 +74,8 @@ export const SupportMessage: FC<SupportMessageProps> = ({
   className,
   description,
   onClick,
-  type = 'info',
+  rightSideComponent,
+  type,
   title,
   ...marginProps
 }) => {
@@ -96,7 +101,8 @@ export const SupportMessage: FC<SupportMessageProps> = ({
           <Description tag="p">{description}</Description>
         )}
       </Box>
-      {onClick && (
+      {rightSideComponent}
+      {onClick && rightSideComponent === undefined && (
         <Box ml={{ custom: 'auto' }}>
           <Icon size={16} render="caret" color="marzipan" rotate={270} />
         </Box>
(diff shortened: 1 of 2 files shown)


src/SupportMessage/storybook/SupportMessage.stories.tsx
```diff
@@ -1,46 +1,62 @@
...
+    title: 'An interactive SupportMessage',
+    rightSideComponent: (
+      <Box ml={{ custom: 'auto' }}>
+        <Link href={undefined as unknown as string}>
+          This is a right side component
+        </Link>
+      </Box>
+    ),
+    description: 'Click me!',
+  },
```
david-gomes5, line 57: Added story example of using new `rightSideComponent` prop

src/SupportMessage/SupportMessage.tsx
```diff
@@ -62,6 +62,10 @@ export type SupportMessageProps = {
   className?: string
   description: string | ReactElement
   onClick?: MouseEventHandler
+  /**
+   * Right side content, usually an icon or a button
+   */
+  rightSideComponent?: ReactElement
```
DeadEnglish, line 68: nit: Do we want to be bit more flexible and use `ReactNode` here? would mean we could use text without wrapping in a component. Would probably mean we want some default flex styling for the parent in these cases (we should probably add this anyway)

DeadEnglish (approved):

src/SupportMessage/SupportMessage.tsx
david-gomes5: The approach I try to have here is two things:
- A approach that favours generic composability over strict extensibility.
- Only adding functionality that is needed, avoiding changes that we're unsure will be used

I don't really have a use case for simply allowing text on the right side, I also have a question on what default props this text fallback would have, making the wrong choice might be more detrimental than beneficial.

On the flip-side, I understand this generic composability approach requires a little more code to take advantage of, which I would say would be the trade-off.

PR #4070, opened 2024-11-01
feat(Tooltip): Refactor tooltip position logic

david-gomes5:
## What does this do?
- Replace existing logic with the `@floating-ui/react` library
  - Arrow position much more responsive
  - Position of tooltip is sticky as things move and resize
- Refactor some props to simplify and future proof the component (includes breaking changes)
- Support adding a tooltip inline. Works with text that spans multiple lines

### BREAKING CHANGES:
- Removed `fallbackStyle` prop, replaced with the `variant` prop (better documentation of which variants are available)
- Removed `tooltipColor` prop, use `variant` instead
- Changed typing of `portalContainer` to accept HTMLElement and similar reference types
- Removed the outer `Container` that wrapped children in a `display: block` element. This might have unintended effects downstream but should improve DX when implementing the component

## Screenshot / Video

https://github.com/user-attachments/assets/712c4c3f-3f05-4af0-83f8-a39bcd0e936b

https://github.com/user-attachments/assets/507c1b2c-f982-4887-83a0-7dbef93e823e

https://github.com/user-attachments/assets/c1f5df3b-75bf-4f3a-914b-77eb05a76e25

> Changes from Storybook

- closes https://www.notion.so/getmarshmallow/Tooltip-77d67a7624bf4c45b6b08748916385db?pvs=4
- closes https://www.notion.so/getmarshmallow/Fix-tooltip-positioning-for-sticky-elements-cc1cc2702e0e4ba59adb568f99ec4358?pvs=4

## Testing
- Using Storybook
- Test in downstream:
  - [x] signup (no usage)
  - [x] agent portal
- [ ] Figure out why [preview](https://github.com/marshmallow-insurance/smores-react/actions/runs/11632250775/job/32394977564#step:5:148) doesn't bump major version


src/Tooltip/Tooltip.tsx
```diff
@@ -1,231 +1,143 @@
...
+  const { refs, floatingStyles, context } = useFloating({
+    open: showTip,
+    placement: position,
+    onOpenChange: setShowTip,
+    middleware: [
+      offset(ARROW_HEIGHT + GAP),
+      flip(),
+      shift(),
+      arrow({ element: arrowRef, padding: 14 }),
+    ],
```
david-gomes5, line 87: [Middleware API](https://floating-ui.com/docs/middleware).

If needed, we can allow for the middleware to be configured at the point of usage.
For now, I'm guessing we don't need that level of configuration.

src/Tooltip/Tooltip.tsx
```diff
@@ -1,231 +1,143 @@
...
 
-  useEventListener({
-    eventName: 'scroll',
-    callback: debounce(handleTipViewport, 100),
-    ref: documentRef,
-  })
+  const hover = useHover(context, { move: false })
+  const focus = useFocus(context)
+  const dismiss = useDismiss(context)
+  const role = useRole(context, { role: 'tooltip' })
```
david-gomes5, line 94: [hover](https://floating-ui.com/docs/useHover): Opens the floating element while hovering over the reference element, like CSS :hover.
[focus](https://floating-ui.com/docs/useFocus): Opens the floating element while the reference element has focus, like CSS :focus.
[dismiss](https://floating-ui.com/docs/useDismiss): Closes the floating element when a dismissal is requested — by default, when the user presses the escape key or outside of the floating element with their pointer.
[role](https://floating-ui.com/docs/useRole): Adds base screen reader props to the reference and floating elements for a given `role`.
  - We could allow this to be configurable but I don't think it'll be needed

src/Tooltip/Tooltip.tsx
```diff
@@ -1,336 +1,192 @@
...
+    open: showTip,
+    placement: position,
+    onOpenChange: setShowTip,
+    middleware: [
+      offset(ARROW_HEIGHT + GAP),
+      flip(),
+      shift(),
+      inlineProp ? inline() : undefined,
+      arrow({ element: arrowRef, padding: 14 }),
+    ],
```
david-gomes5, line 105: [Middleware API](https://floating-ui.com/docs/middleware).

If needed, we can allow for the middleware to be configured at the point of usage.
For now, I'm guessing we don't need that level of configuration.

matdiak (approved):

PR #4076, opened 2024-11-05
feat(TextDateInput): add text date input component

ReginaGyimesi:
## What does this do?
Moved the date of birth input component over from the signup flow to s'mores as there was a need to use it in CMS, too. 
The component also contains validation to ensure we don't leave room for ambiguity when checking the dates. 

## Screenshot / Video
Screenshot 2024-11-07 at 14 42 37
Screenshot 2024-11-07 at 14 42 51

[RFC](https://www.notion.so/getmarshmallow/Date-of-birth-input-component-bdd970970955456e94d44429f76fcbaf)

## Testing
The PR contains unit tests written by Mat. Can also be manually tested by entering inputs.


src/TextDateInput/TextDateInput.tsx
```diff
@@ -0,0 +1,131 @@
...
+  label,
+  assistiveText,
+  error,
+  showCompleted = false,
+}: TextDateInputProps) => {
+  const [errorMsg, setErrorMsg] = useState<string | undefined>(undefined)
+  const [fieldsWithError, setFieldsWithError] = useState<string[]>([])
+
+  const handleValidationErrors = () => {
+    const validationResult = dateOfBirthValidator(value)
```
david-gomes5, line 38: I would argue, because of this validator, this component isn't a generic `TextDateInput` but a more opinionated `DateOfBirthDateInput` component.

I think we'll need to decide how to handle configurable validation before this component can be generic enough to be named `TextDateInput`.

Example:
- One would assume this component is usable for all past dates but the validator doesn't allow for dates less than 17 years ago
- One would assume this component is usable for future dates but doesn't

david-gomes5 (changes requested):

src/TextDateOfBirthInput/TextDateOfBirthInput.tsx
```diff
@@ -24,7 +24,7 @@ export type TextDateOfBirthInputProps = {
 } & Pick<FieldsetProps, 'label' | 'error' | 'assistiveText'>
 
 /**
- * This component renders a set of input fields to collect a date of birth as separate day, month,
+ * Renders a set of input fields to collect a date of birth as separate day, month,
  * and year fields. It handles validation using `dateOfBirthValidator`, which checks for invalid formats,
  * impossible values (e.g., days or months out of range), missing fields, and specific constraints
  * (e.g., age must be at least 17 years, year cannot be before 1900).
```
david-gomes5, line 30: Looking good, this suggestion is nitpicky (sorry) but jsdoc supports markdown so you could list your examples with bullet points:
```tsx
/**
*  e.g.
* - Age must be at least 17 years
* - Selected year cannot be before 1900
* /
```

Moving these items into a list might be easier to scan when a developer is hovering over the component to check its docs

src/TextDateOfBirthInput/TextDateOfBirthInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Renders a set of input fields to collect a date of birth as separate day, month,
+ * and year fields.
+ *
+ * @function dateOfBirthValidator validates a date of birth provided in a DateObject format:
+ * - All fields are present and correctly formatted.
+ * - The age is at least 17 years old.
+ * - The year is not before 1900.
+ */
+export const TextDateOfBirthInput = ({
```
matdiak, line 35: I can see that this is no longer `forwardRef`, I think we need that in order to be able to reuse this in sign-up. 

Treating is as a ref also allows for react-hook-form to then focus the field in case of a validation error.

src/TextDateOfBirthInput/TextDateOfBirthInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+      setFieldsWithError(validationResult.fields || [])
+    } else {
+      setErrorMsg(undefined)
+      setFieldsWithError([])
+    }
+  }
+
+  useEffect(() => {
+    handleValidationErrors()
+  }, [value])
```
matdiak, line 64: Correct me if I am wrong but the component but you are planning to use zod in order to implement this?

If so, I think this logic should sit outside of the component and the component should be stateless, taking the state from react hook form. To clarify, I think `dateOfBirthValidator` should be use within the schema.

src/TextDateOfBirthInput/storybook/TextDateOfBirthInput.stories.tsx
david-gomes5, line 1: Apologies, realised this late - could we migrate to using Storybook's new [CSF](https://storybook.js.org/docs/api/csf)?

Some components' stories have already been migrated like `Button.stories.ts`

src/TextDateOfBirthInput/TextDateOfBirthInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+export type DateObject = {
+  [K in keyof DateObjectType]?: DateObjectType[K] | null
+}
+
+export type TextDateOfBirthInputProps = {
+  value: DateObjectType
+  onChange: (value: DateObjectType) => void
+  showCompleted?: boolean
+  fieldsWithError?: fieldsWithErrorType[]
```
matdiak, line 24: Why is this prop now needed?

package.json
```diff
@@ -92,7 +92,8 @@
     "fuse.js": "^7.0.0",
     "lexical": "^0.19.0",
     "micromatch": "^4.0.8",
-    "polished": "^4.1.3"
+    "polished": "^4.1.3",
+    "react-hook-form": "^7.53.2"
```
matdiak, line 96: How come did you decide to install the library?

I think you can get away with just typing the hookformError as
```
{
    type: string
    message?: string
  }
```

src/TextDateOfBirthInput/TextDateOfBirthInput.tsx
```diff
@@ -4,25 +4,26 @@ import styled from 'styled-components'
...
 import { Fieldset, FieldsetProps } from '../fields/Fieldset'
 import { TextInput } from '../TextInput'
+import { InputValidationError } from '../utils/dateOfBirth/schema'
+import { FieldError } from 'react-hook-form'
 
-export interface DateObjectType {
-  day: string
-  month: string
-  year: string
+export interface CompleteDateObject {
```
matdiak, line 10: Nitpick: Is this type even used now that we have `DateObject`?

I think if you are not using it in cms then you can either:
- delete it and make 
```
export type DateObject = {
  day?: string | null
  month?: string | null
  year?: string | null
}
```
- or you can just keep each key of `CompleteDateObject` type as non-undefined as the DateObject makes them optional anyways `[K in keyof CompleteDateObject]?:`

matdiak (approved):
I think all good. Thanks for patience. 

I think it would be good not to introduce the library as a dependency but I don't want to delay you so approved.

PR #4158, opened 2025-01-02
feat: export date of birth validator utils

matdiak:
## What does this do?
In process of refactoring sign-up to use the smores `TextDateOfBirthInput`, I realised that `getImpossibleValues` util is used in sign-up flow so exporting it. 

## Screenshot / Video
n/a

n/a

## Testing
n/a


DeadEnglish (approved):
We could potentially move this to Campfire in the near future 👀

PR #4170, opened 2025-01-09
fix(dateOfBirthValidator): Unable to build downstream apps because of missing import

david-gomes5:
[INT-68]

## What does this do?
- Fix unable to build downstream apps because of missing import

When building out package assets, the asserts' imports aren't coerced into relative imports which break namespace imports like the one found in this PR.

## Screenshot / Video
Screenshot 2025-01-09 at 11 15 11

## Testing
Downstream

[INT-68]: https://marshmallow1.atlassian.net/browse/INT-68?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/dateOfBirth/dateOfBirthValidator.ts
+++ b/src/utils/dateOfBirth/dateOfBirthValidator.ts
@@ -9,7 +9,7 @@ import {
 } from './dateOfBirthValidator.utils'
 import { FieldValidationError, InputValidationError } from './schema'
 
-import { DateObject } from 'TextDateOfBirthInput'
+import { DateObject } from '../../TextDateOfBirthInput'
 
 export const dateOfBirthValidator = (dateObject: DateObject) => {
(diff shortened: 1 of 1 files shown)


DeadEnglish (approved):

matdiak:
@david-gomes5 can I ask how is this fixing the issue?

david-gomes5:
> @david-gomes5 can I ask how is this fixing the issue?

I added more description to my PR late but basically:
> When building out package assets, the asserts' imports aren't coerced into relative imports which break namespace imports like the one found in this PR.

In the downstream, when importing packages in `node_modules`, it can't do much with 'TextDateOfBirthInput' as an import path because it doesn't know where to look for that, it probably looks in the root of `node_modules` but doesn't find anything.

PR #4171, opened 2025-01-09
fix(React): Fix types after react 19 upgrade

david-gomes5:
[INT-69]

## What does this do?
The [React 19 upgrade](https://github.com/marshmallow-insurance/smores-react/pull/4166) has caused types to error. PR here is to fix erroring types

These changes shouldn't be breaking as it should still work in the downstream apps that have older versions of React types (ones that don't have react 19 or 18.3 installed).

> [!NOTE] 
> These changes aren't breaking for downstream apps that haven't upgraded to React 19 yet as the inputs of the changed types haven't changed. Tested on `agent-portal-www` and found no issue.

## Screenshot / Video
Only typing changes

[INT-69]: https://marshmallow1.atlassian.net/browse/INT-69?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@storybook/addon-essentials": "^8.4.7",
     "@storybook/addon-interactions": "^8.4.7",
     "@storybook/addon-links": "^8.4.7",
+    "@storybook/preview-api": "^8.4.7",
     "@storybook/react": "^8.4.7",
     "@storybook/react-vite": "^8.4.7",
     "@storybook/test": "^8.2.1",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -86,7 +87,6 @@
   "dependencies": {
     "@floating-ui/react": "0.27.3",
     "@lexical/react": "^0.23.0",
-    "@storybook/preview-api": "^8.4.7",
```
david-gomes5, line 89: Moving this out of dependencies as we shouldn't have downstream apps install this as it is a dev dep

src/Badge/Badge.tsx
```diff
@@ -18,7 +18,7 @@ export type BadgeProps = {
   zIndex?: number
 }
 
-export function Badge<T extends BadgeProps>({
+export function Badge({
```
david-gomes5, line 21: Unused generic that doesn't actually get used to enrich any types

matdiak (approved):

PR #4216, opened 2025-01-29
feat: upgrading eslint to v9 and removing unneeded deps

rowanhen:
## What does this do?

- upgrades eslint to 9.19.0
- removes unused babel dependencies
- moves .eslintrc to use eslint.config.js
- moves .lintstagedrc to use lint-staged.config.js
- moves .prettierrc to use prettier.config.js
- addresses most eslint associated issues 

> [!TIP]
>  review with whitespace hidden

- closes: https://github.com/marshmallow-insurance/smores-react/pull/4209


eslint.config.js
rowanhen, line 1: pretty much 1-1 with Frames / AP now

eslint.config.js
```diff
@@ -0,0 +1,61 @@
...
+      '@typescript-eslint/no-unsafe-argument': 'off',
+      '@typescript-eslint/no-floating-promises': 'off',
+      // TODO: Temporary warn, to fix in separate PR
+      '@typescript-eslint/no-unsafe-return': 'warn',
+      // TODO: Temporary warn, to fix in separate PR
+      '@typescript-eslint/no-explicit-any': 'warn',
+      '@typescript-eslint/prefer-nullish-coalescing': 'warn',
+      strict: ['error', 'never'],
+      'react-hooks/rules-of-hooks': 'error',
+      'react-hooks/exhaustive-deps': 'off',
```
david-gomes5, line 58: Would we be able to warn here?

Logs might be noisy for ESLint but at least it'll surface in the IDE during development to guide the developer to resolve.

package.json
```diff
@@ -17,6 +18,7 @@
     "build": "npm run clean && tsc --p ./tsconfig.build.json",
     "build-storybook": "storybook build -c .storybook",
     "check-types": "tsc --noEmit",
+    "lint": "eslint .",
```
david-gomes5, line 21: Should it be `eslint . --fix`?

src/LabelledText/LabelledText.tsx
```diff
@@ -8,7 +8,7 @@ import { Text } from '../Text'
 export type LabelledTextProps = {
   children: ReactNode
   label: string | JSX.Element
-  labelMargin?: 2 | 4 | 8 | number
+  labelMargin?: number
```
david-gomes5, line 11: I think we should probably disable the ESLint rule that caused this for smores.

I think from a DX prospective, it might be really nice to have autocomplete suggestions while still being able to set the value to any number

src/SegmentedControl/ToggleButton.tsx
```diff
@@ -22,7 +22,7 @@ export const ToggleButton = <Value,>({
   return (
     <Wrapper
       as="label"
-      data-testId={`value-toggle-${value}`}
+      data-testId={`value-toggle-${value as string}`}
```
david-gomes5, line 25: Would it work if we coerce it to a string instead? underneath this template literal, it should call something similar to `toString` or String.

```suggestion
      data-testId={`value-toggle-${String(value)}`}
```

src/Table/components/TableRow.tsx
```diff
@@ -119,7 +117,7 @@ export const TableRow = <T extends object>({
               {
                 // TODO: React 19 requires the props type by default. It was previously inferred to be 'any'. It would be beneficial to explicitly type this generic value
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
david-gomes5, line 119: Are we able to remove the comments now or no?

src/Table/components/commonComponents.tsx
```diff
@@ -24,7 +24,7 @@ export const StyledHeaderCell = styled.th<TransientProps<TableStylesProps>>`
   z-index: 2;
   text-align: left;
   vertical-align: bottom;
-  ${fontStyleMapping['label']};
+  ${fontStyleMapping.label};
```
david-gomes5, line 27: ESLint rules like these are 🔥

src/Table/components/RowActions.tsx
```diff
@@ -30,59 +30,64 @@ export const RowActions = <T extends object>({
...
+                    className: 'reactElementRowAction',
+                  })}
+                {!('element' in action) && action.genericButton && (
+                  <Button
+                    {...action.genericButton}
+                    handleClick={(e) => {
+                      void (async () => {
                         await handleAction(e, action.onClick)
-                      }}
+                      })()
```
david-gomes5, line 52: I think instead of doing this we can just remove the async from the callback and prefix the line with void like so:
```
handleClick={(e) => void handleAction(e, action.onClick)}
```

I actually don't think I should've turned these into async in the first place, there was a reason for it but I think I didn't need to add it here.
This essentially makes it explicit that we don't care and won't use the return value of the callback

david-gomes5: Looking at the [work](https://github.com/marshmallow-insurance/smores-react/pull/4025/files) that added this, I think I just needed to change the type for allowing for passing a promise but I don't think it was necessary to add all the awaits, there was no reason for it.

david-gomes5 (changes requested):

david-gomes5 (approved):

PR #4218, opened 2025-01-30
chore: adding strict mode to tsconfig

rowanhen:
## What does this do?

- adds strict mode to tsconfig
- addresses associated errors - was pretty much just all our .stories files - so i took the chance to convert them to the new storybook boilerplate in which we can extend them in the future (recommended boiler plate here - https://storybook.js.org/docs/writing-stories)

**Other unrelated changes**
- deletes `BrandCard` & `Emoji` components - not being used anywhere


tsconfig.json
```diff
@@ -3,30 +3,16 @@
     "outDir": "./dist",
     "module": "esnext",
     "target": "esnext",
-    "lib": [
-      "es5",
-      "es6",
-      "es7",
-      "es2017",
-      "es2019",
```
rowanhen, line 11: not supported in any repos we use rn (other than CA but that dm) - so just esnext is probably fine here

tsconfig.json
```diff
@@ -3,30 +3,16 @@
...
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["esnext", "dom", "dom.iterable"],
     "types": ["react", "node", "vitest/globals"],
     "sourceMap": true,
     "inlineSources": true,
-    "allowJs": true,
-    "checkJs": false,
```
rowanhen, line 20: we have no js in our src folder anymore so can delete

tsconfig.json
```diff
@@ -3,30 +3,16 @@
...
     "jsx": "react",
     "declaration": true,
     "moduleResolution": "Bundler",
     "baseUrl": "src",
     "forceConsistentCasingInFileNames": true,
-    "noImplicitReturns": true,
-    "noImplicitThis": true,
-    "noImplicitAny": true,
-    "strictNullChecks": true,
```
rowanhen, line 29: deletes these in favour of strict

david-gomes5:
Since this is a public component library and I'm presuming we follow semvar, then I think we should flag this one as a breaking change. As we're deleting two components.

src/Accordion/storybook/Accordion.stories.tsx
```diff
@@ -1,18 +1,64 @@
...
 import React from 'react'
-import { Accordion, AccordionProps } from '../Accordion'
-import { CollectionPage } from './Collection'
+import { Box } from '../../Box'
+import { Accordion } from '../Accordion'
 
-export default {
-  title: 'Accordion',
+const meta: Meta<typeof Accordion> = {
+  title: 'Accordian',
```
david-gomes5, line 7:
```suggestion
  title: 'Accordion',
```

david-gomes5 (approved):
🔥

PR #4230, opened 2025-02-04
BREAKING CHANGE: deleting BrandCard and Emoji components, adjusting package type to esm

rowanhen:
## What does this do?

PR to bump for a Major version for the following breaking changes:

1. Deleting `Emoji` and `Brandcard` components (unused internally)
2. Adjusting package.json to type: "module", this will require some changes for our components using Vitest

Steps to fix (2.)

Add the following to vitest.config - 

```
      deps: {
        inline: ['@mrshmllw/smores-react'],
      },
```


src/Icon/iconsList.ts
rowanhen, line 1: formatted by ascending

david-gomes5 (approved):

PR #4263, opened 2025-02-21
feat: removes deprecated typos

matdiak:
BREAKING CHANGE: removes deprecated typos

## What does this do?

Removes deprecated typos from s'mores. This will be released as a major version update.

Changes old | new
Screenshot 2025-02-24 at 11 30 00
Screenshot 2025-02-24 at 11 32 04
Screenshot 2025-02-24 at 11 33 01
db39a2c-96b4-4468-be9d-f0a1d49e5b6b" />
Screenshot 2025-02-24 at 11 35 02

## Screenshot / Video

n/a

https://github.com/orgs/marshmallow-insurance/projects/7/views/58?pane=issue&itemId=98919346&issue=marshmallow-insurance%7Cgrowth%7C2442

## Testing

n/a


src/Text/Text.tsx
```diff
@@ -97,149 +79,14 @@ export const Text: FC<TextProps> = forwardRef<HTMLElement, TextProps>(
 
 Text.displayName = 'Text'
 
-const isNewTypo = (value: string): value is Typo => {
+const isTypo = (value: string): value is Typo => {
   return Object.keys(fontStyleMapping).includes(value)
 }
 
 const Container = styled(Box)<IText>(
-  ({ $align, $color, $cursor, $typo }) => css`
```
david-gomes5, line 105: Is there a reason we're moving `$align`, `$color` and `$cursor`?

david-gomes5 (changes requested):
Would we be able to strictly type the `typo` prop with the `Typo` type?

rowanhen:
to make sure you trigger a breaking change - its worth pushing a commit message using `--no-verify` to bypass the commitlint in this format: `BREAKING CHANGE: removes deprecated typos`, along with renaming the PR title to something similar

matdiak:
> Would we be able to strictly type the typo prop with the Typo type?
@david-gomes5 the type `string &Record<never, never>` is used so that we get string literal auto-suggestions ([ref](https://github.com/microsoft/TypeScript/issues/29729#issuecomment-567871939))

src/Text/Text.tsx
matdiak: That's a mistake :(

matdiak: Fixed now

david-gomes5:
> > Would we be able to strictly type the typo prop with the Typo type?
> 
> @david-gomes5 the type `string &Record<never, never>` is used so that we get string literal auto-suggestions ([ref](https://github.com/microsoft/TypeScript/issues/29729#issuecomment-567871939))

Would we not be able to achieve auto-suggestions by simply using the `Typo` type?

matdiak:
> Would we be able to strictly type the typo prop with the Typo type?

@david-gomes5 the type string &Record<never, never> is used so that we get string literal auto-suggestions (https://github.com/microsoft/TypeScript/issues/29729#issuecomment-567871939)

[matdiak](https://github.com/matdiak) Would we not be able to achieve auto-suggestions by simply using the Typo type? I.e dropping the string

I was wrong it seems. It does work just with Typo. I thought I tried to quickly and it didn't but it does now so I think we can simplify the type to just Typo.

src/Text/Text.tsx
```diff
@@ -38,7 +38,7 @@ type Props = {
   className?: string
   // Record<never, never> matches any non-null and non-undefined type so
   // IntelliSense can suggest Typo to autocomplete
```
david-gomes5, line 40: This comment is now obsolete, would we be able to remove it?

david-gomes5 (approved):

matdiak:
@rowanhen I addressed the major update issue in [this PR](https://github.com/marshmallow-insurance/smores-react/pull/4268)

PR #4268, opened 2025-02-24
fix: add breaking change config to semantic release

matdiak:
## What does this do?

In https://github.com/marshmallow-insurance/smores-react/pull/4263#issuecomment-2674951765, we discussed how we need to override the conventional commit naming in order for semantic release to identify a breaking change as a major release.

This PR adds a config based on [default rules](https://github.com/semantic-release/commit-analyzer/blob/master/lib/default-release-rules.js) in order to respect the breaking change in the `releaseRules` order.

Example of Github action flows:
Recognised major update to v12 by presence of `BREAKING CHANGE: ` in the footer - [ref](https://github.com/marshmallow-insurance/smores-react/actions/runs/13498114124/job/37709851380)
Without the config change
https://github.com/marshmallow-insurance/smores-react/actions/runs/13497862202/job/37709103463

## Screenshot / Video

n/a

n/a

## Testing

n/a

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -18,6 +18,7 @@ const config = {
       {
         preset: 'angular',
         releaseRules: [
+          { breaking: true, release: 'major' },
           { type: 'feat', release: 'minor' },
           { type: 'fix', release: 'patch' },
           { type: 'docs', release: 'patch' },


rowanhen:
as apart of this, do you mind adjusting our commitlint config to what has been added here - https://github.com/marshmallow-insurance/campfire/blob/main/libs/configs/commitlint.config.ts

this just adds the ignore rule when we want to use a `breaking change: XYZ` commit message so we dont need to use --no-verify everytime

david-gomes5 (approved):
🙌

matdiak:
> as apart of this, do you mind adjusting our commitlint config to what has been added here - https://github.com/marshmallow-insurance/campfire/blob/main/libs/configs/commitlint.config.ts   
this just adds the ignore rule when we want to use a breaking change: XYZ commit message so we dont need to use --no-verify everytime

@rowanhen do you mean to add the ignore rule to allow "breaking change: " in the commit? I think the convention is that you keep the conventional commits but specify the breaking change in the footer ([ref](https://github.com/semantic-release/semantic-release?tab=readme-ov-file#how-does-it-work), [ref](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#commit-message-footer))

matdiak:
```
commit 0dbc26e516ea8572c1ff8f8bfdbe1a18f0a3eaec
Author: Mat Diak <mat.diak@marshmallow.co>
Date:   Fri Feb 21 17:07:05 2025 +0000

    feat: remove deprecated typos (#4264)
    
    * feat: removes deprecated typos
    
    * fix: revert removing other props
    
    BREAKING CHANGE: removes deprecated typos
```

So your commit would look like so btw.

PR #4334, opened 2025-03-26
fix: adding SMORES_INTERNAL_FIELD classname to all fields

rowanhen:
## What does this do?

adds `SMORES_INTERNAL_FIELD` classname to all fields for ease of styling and finding smores fields on a single page


DeadEnglish:
I think this change raises a few accessibility concerns. `fieldset` should be used to group a section of components, having every input belong to it's own `fieldset` would probably introduce some issues e.g.

- If all inputs are grouped in their own fieldset, we're not really using fieldset for its intended purpose
- `fieldset` and `legend` are read out by screenreaders, this could be somewhat annoying for ever field
- Possibly styling changes?
  - I haven't checked what default styling we have for `fieldset` but it might not be what we want
  
Also in this PR we're changing the wrapper to a `fieldset` and we would also need to use `legend` for the input and I don't think we'd want to do that for every input we have.

rowanhen:
@DeadEnglish fair points - have, adjusted this so that we instead just add a generic className instead

PR #4364, opened 2025-04-08
feat(progress-indicator): new component

DeadEnglish:
## What does this do?
A new progress indicator component to show progress within form steps. This has the following props

- `steps`: An array of objects with the following properties
  - `handleStepChange`: A function that runs on step change, does not run if the step is `disabled` (A step is considered `disabled` if it's not the current step or a completed step)
  - `onStepClick`: A function that runs on step click. This is always run regardless of step state
  - `simpleStep`: sets the component to a "simple" view (see below screenshots)
  - `currentStep`: the current step the form should be on

> [!NOTE]
Whilst this is only used in the signup flow, this is part of the components library so we don't have any deviation going forward.

### Normal progress indicator
Screenshot 2025-04-08 at 15 22 03

https://github.com/user-attachments/assets/2c4a31c1-41d1-4a3d-a135-b4e01fce555d

### Simple progress indicator
Screenshot 2025-04-08 at 15 29 11

https://github.com/user-attachments/assets/eee6d8c5-ece5-4354-a718-64338b4ff4c9

- [Component within design system](https://www.figma.com/design/FqSaizGOyOzXFZNmZ4X0LGMn/%F0%9F%8D%AC-S-mores?node-id=6862-27666&m=dev&focus-id=21455-112418)

## Testing
- New unit tests added for each component

### DIFF
--- a/src/ProgressIndicator/components/StepItem.tsx
+++ b/src/ProgressIndicator/components/StepItem.tsx
@@ -3,26 +3,29 @@ import { StepData } from 'ProgressIndicator/types'
 import React from 'react'
 import styled, { css } from 'styled-components'
 import { theme } from '../../theme'
+import { Text } from '../../Text'
+import { Icon } from '../../Icon'
 
-export interface Props extends Pick<StepData, 'label' | 'isHidden'> {
+export interface StepItemProps extends Pick<StepData, 'label'> {
   isCompleted?: boolean
   isLastCompleted?: boolean
+  isCurrentStep: boolean
   isSimple?: boolean
   onClick: () => void
   stepWidth: string
+  isLastItem: boolean
 }
 
 export const StepItem = ({
-  isHidden,
+  label,
   isSimple = false,
+  isCurrentStep,
   stepWidth,
+  onClick,
   isCompleted = false,
   isLastCompleted = false,
-}: Props) => {
-  if (isHidden) {
-    return null
-  }
-
+  isLastItem = false,
+}: StepItemProps) => {
   if (isSimple) {
     return (
       <SimpleItem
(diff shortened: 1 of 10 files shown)


src/ProgressIndicator/storybook/ProgressIndicator.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+            'Boolean used to determine if we should render the simple step view. Note: the simple step layout has limited functionality e.g. No onClick function',
+        },
+        defaultValue: {
+          summary: 'false',
+          detail: 'defaults to false and will show the normal step layout',
+        },
+      },
+    },
+
+    // simpleStep: { description: 'used to determine ', 'type' },
```
HassanTAKhan, line 74: Do we need this?

src/ProgressIndicator/helpers.ts
```diff
@@ -0,0 +1,33 @@
...
+interface ReturnValue {
+  totalWidth: number
+  stepWidth: number
+}
+
+export const calulateProgressWidths = (
+  stepCount: number,
+  isSimpleSteps?: boolean,
+): ReturnValue => {
+  // if not simple steps, we need to subtract 1 from the step count as the last item doesn't have a line
```
HassanTAKhan, line 14: Do we need this?

src/ProgressIndicator/components/StepItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+      onClick={onClick}
+    >
+      <ProgressIndicator
+        $completed={isCompleted}
+        $currentStep={isCurrentStep}
+        flex
+        alignItems="center"
+        justifyContent="center"
+      >
+        {isCompleted ? <Icon render="tick" size={16} color="cream" /> : null}
```
HassanTAKhan, line 55: I feel like `{isCompleted && <Icon render="tick" size={16} color="cream" />}` would be more readable

src/ProgressIndicator/components/StepItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+      </ProgressIndicator>
+      {isCompleted && !isLastItem ? <CompletedBar /> : null}
```
HassanTAKhan, line 57: Same as above

src/ProgressIndicator/helpers.ts
DeadEnglish: I've kept this so we know why we're doing it, otherwise it seems like an odd thing to do 😅

HassanTAKhan (approved):

PR #4413, opened 2025-05-07
feat(tag): added optional onclick prop to tag comp

geriszabo:
## What does this do?
Added optional onClick property for Tag componnet

## Screenshot / Video
image

### DIFF
--- a/src/Tag/Tag.tsx
+++ b/src/Tag/Tag.tsx
@@ -17,6 +17,7 @@ export type TagProps = {
   className?: string
   icon?: Icons
   iconColor?: Color
+  onClick?: () => void
 } & MarginProps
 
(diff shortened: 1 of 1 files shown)


david-gomes5 (approved):

david-gomes5:
@geriszabo would it make sense to showcase this new option in Storybook? I'm unsure if it surfaces intrinsically already

geriszabo:
> @geriszabo would it make sense to showcase this new option in Storybook? I'm unsure if it surfaces intrinsically already

i mean its a native property for all html elements, i dont know if it would make sense to showcase it. I have an edge case with this component and thats why i need it, never seen a Tag component being clicked elsewhere in the repo 😅

PR #4419, opened 2025-05-08
feat(radiogroup): radiogroup and dropdown style

geriszabo:
## What does this do?
Changed the styling of the radiogroup if the option is disabled, changed font color of placeholder text

## Screenshot / Video
image

image

PR is part of this ticket: https://marshmallow1.atlassian.net/browse/CLM-1364

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -181,13 +181,15 @@ interface SSelect {
   $error: boolean
   $frontIcon?: string
   $fallbackStyle?: boolean
+  value?: string | null
 }
 
 const StyledSelect = styled.select<SSelect>`
   ${resetSelect}
   width: 100%;
   height: 32px;
-
+  color: ${({ value }) =>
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -126,7 +126,7 @@ export const Dropdown = forwardRef(function Dropdown(
           name={name}
           $frontIcon={frontIcon}
           $fallbackStyle={fallbackStyle}
-          value={value ? value : ''}
+          value={value ?? ''}
```
adam-magyar, line 129: nice catch! 
I would suggest to give value a default value as we desctructure it from the props like:
```javascript
value: valueProp = '',
```

and then we can just pass value here
```suggestion
          value={value}
```

adam-magyar (approved):
dropped a minor comment, at all LGTM 👍

src/Dropdown/Dropdown.tsx
geriszabo: nice catch as well!:D gave it a default value so now it looks cleaner

PR #4423, opened 2025-05-09
feat(dropdown): fix dropdown controlled to uncontrolled

geriszabo:
## What does this do?
Fixes bug where dropdown was switched to controlled

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -58,7 +58,7 @@ export const Dropdown = forwardRef(function Dropdown(
     showDefaultOption = false,
     customDefaultOption,
     name,
-    value: valueProp = '',
+    value: valueProp,
     defaultValue,
     disabled = false,
     list,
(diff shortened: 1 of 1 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -126,7 +126,7 @@ export const Dropdown = forwardRef(function Dropdown(
           name={name}
           $frontIcon={frontIcon}
           $fallbackStyle={fallbackStyle}
-          value={value}
+          value={value ?? ''}
```
david-gomes5, line 129: This should still cause the element to be controlled by react.

The value prop has to be `undefined` to be uncontrolled — what was the original aim of defaulting to an empty string?

geriszabo: i dont quite know what the original aim was but this was the original code yesterday:
https://github.com/marshmallow-insurance/smores-react/pull/4419/commits/dbabfc55c7ca35bd975d069307eae7db1c8f650b

snegyeliczky:
As we have the discussion about the issue I approve

snegyeliczky (approved):

david-gomes5:
This PR doesn't fix the issue.

@snegyeliczky / @geriszabo

PR #4427, opened 2025-05-12
fix(Table): Fix custom table padding

david-gomes5:
Fixes custom table padding as the default padding was overriding the custom values


src/Table/types.ts
```diff
@@ -124,9 +124,9 @@ interface CommonTableProps<T> {
   /** Array of actions that can be performed on each row. */
   rowActions?: RowActions<T>
   /** The Y padding for each row. */
-  rowPadding?: string
+  rowPadding?: SingleSpacing
```
david-gomes5, line 127: Ensure we're only passing single pixel value instead of multiple values

e.g. 
```
18px
```

not
```
18px 18px 18px
```

sapesz (approved):
nice! ☺️

PR #4537, opened 2025-06-17
chore(dependabot): add storybook group for dependency updates

david-gomes5:
group them up to reduce the maintenance burden


.github/dependabot.yml
```diff
@@ -40,6 +48,3 @@ updates:
         update-types:
           - "minor"
           - "patch"
-        update-types:
-          - "minor"
-          - "patch"
```
david-gomes5, line 45: duplicate field, woops

rowanhen (approved):

PR #4541, opened 2025-06-27
feat(table): add optional  for row identification in Table component

david-gomes5:
Adding typing for  to avoid cases where we reuse a cell component to transition to another cell's state because they share the same cell index. This is a NON-BREAKING change as the new  is optional, we fallback to using rowIndex

### DIFF
--- a/src/Table/Table.tsx
+++ b/src/Table/Table.tsx
@@ -6,6 +6,17 @@ import { StyledCell, StyledTable } from './components/commonComponents'
 import { TableProps } from './types'
 import { TableFooter } from './components/TableFooter'
 
+type RowData = Object & {
+  /**
+   * It is recommended to have a rowKey for each row, but it is not mandatory.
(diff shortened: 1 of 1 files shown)


DeadEnglish (approved):

src/Table/types.ts
```diff
@@ -131,6 +131,20 @@ interface CommonTableProps<T> {
...
+/**
+ *  The type of data that each row in the table should conform to.
+ */
+export type TableRowData = {
+  /**
+   * It is recommended to have a rowKey for each row, but it is not mandatory.
+   * This key is used to uniquely identify each row in the table to accurately reconcile each row between React renders.
+   *
+   * If not provided, the row index will be used as the key.
+   * This might not ideal as unrelated rows may cause unexpected transitions if they change position within the table.
```
david-gomes5, line 143:
```suggestion
   * This might not be ideal as unrelated rows may cause unexpected transitions if it changes position within the table.
```

PR #4567, opened 2025-09-30
feat(progress-indicator): align with design system [DES-67]

DeadEnglish:
## What does this do?
- Uses theme provider colours for progress indicator

> [!NOTE]
> Design have okayed the colour change 🎉  

### Left - prod | right - dev
https://github.com/user-attachments/assets/157497ec-1942-42f5-a998-f30a08ff40a0

N/A

## Testing
- snapshot tests added
- snapshot tests updated


david-gomes5 (approved):
Approving with the expectation that we will verify colour change

PR #5105, opened 2026-08-10
fix(SearchInput): reset results scroll position when search query changes

ajayshah95:
## Summary
- `SearchOptions` (used by `SearchInput`'s searchable dropdowns) kept its previous scroll offset when a new search query produced a different filtered list, so the highest-relevance matches could be hidden below the fold until the user manually scrolled back up.
- Adds a `useEffect` keyed on the search query that resets the results list `scrollTop` to `0` and `highlightedIndex` to `-1` whenever the query changes, without touching the existing keyboard-navigation `scrollIntoView` behavior.

## Test plan
- [x] New test: typing a query while scrolled down resets scroll to top
- [x] Existing behavior verified: arrow-key navigation still scrolls the highlighted item into view
- [x] Existing behavior verified: selecting an option still closes the list and resets state
- [x] Full suite (`npm test`), `check-types`, and `eslint` pass with no regressions

### DIFF
--- a/src/SearchInput/SearchInput.test.tsx
+++ b/src/SearchInput/SearchInput.test.tsx
@@ -1,0 +1,63 @@
+import { beforeEach, expect, it, vi } from 'vitest'
+import { fireEvent, render, screen } from '../testUtils'
+import { noop } from '../utils/noop'
+import { SearchInput } from './SearchInput'
+
+const searchList = Array.from({ length: 30 }, (_, i) => ({
+  label: `Option ${i}`,
+  value: `option-${i}`,
+}))
+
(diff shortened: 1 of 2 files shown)


david-gomes5:
@ajayshah95 Would it be possible to see a QuickTime recording of the change?

ajayshah95:
> @ajayshah95 Would it be possible to see a QuickTime recording of the change?

Hey @david-gomes5 - here's a recording which hopefully explains the issue I'm trying to solve.


https://github.com/user-attachments/assets/97b85b18-4eaf-4ee6-b164-3882876a1e88

david-gomes5 (approved):
Nice 👍

PR #5118, opened 2026-08-14
chore: migrate to TypeScript 7 and oxlint [migrate-ts7 dry run]

david-gomes5:
## What does this do?

Takes smores-react from TypeScript 5.9.3 to 7.0.2, ESLint to oxlint, and campfire 6.1.4 to 7.1.0.

**This started as a dry run of the [`migrate-ts7` skill](https://github.com/marshmallow-insurance/claude-code-plugins/pull/190), not as a change intended to land.** The migration itself works and every check is green, but it is opened as a draft because two decisions are outstanding (below). Read it as a working proof that the route is viable, and as the evidence behind the skill changes in that PR.

Impact is global — it changes the compiler, the linter and the published output of a package every frontend consumes.

**The four commits are the phase structure and are worth reviewing in order.** Each was verified against baseline before the next was started, and each is independently revertable:

| Commit | Phase |
|---|---|
| `b7477b7b` | campfire 6.1.4 → 7.1.0 |
| `c9a984d7` | ESLint → oxlint, extending campfire's shared config |
| `899d0da9` | TypeScript 5.9.3 → 6.0.3 |
| `515bbc52` | TypeScript 6.0.3 → 7.0.2 |

A real migration would ship these as four stacked PRs rather than one branch; they are together here because this was a single exploratory run.

### Linting

The oxlint config now extends campfire's shared React config instead of restating rules locally. Two of campfire's rules are switched off deliberately:

```ts
// smores-react DEFINES the color prop and theme colors these rules police in consumers
'campfire/no-color-prop': 'off',
'campfire/no-theme-colors': 'off',
```

Those rules exist to police *consumers* of smores-react, so pointed at this repo they flagged 39 violations against the library that defines the APIs.

`lint-staged.config.js` stops re-exporting campfire's shared config and runs `oxlint` directly. Campfire's version still runs `eslint`, which after this migration resolves transitively, finds no flat config, and exits **0** — so pre-commit would silently stop linting. That needs a separate fix in campfire.

### Published output

TypeScript 7's emit is not byte-identical, which matters here because `dist` is the deliverable. 5.9.3 → 6.0.3 was byte-for-byte identical; 6.0.3 → 7.0.2 changed **123 of 1191 files** in three cosmetic classes:

| Change | Example | Files |
|---|---|---|
| `$`-prefixed JSX props no longer quoted | `"$size": x` → `$size: x` | 55 `.js` |
| `jsx-runtime` specifiers sorted | `{ jsx, jsxs, Fragment }` → `{ Fragment, jsx, jsxs }` | (same files) |
| `.d.ts` union members sorted (`stableTypeOrdering`) | `"$m" \| "$mx"` → `"$alignContent" \| "$alignItems"` | 12 `.d.ts` |

All three are semantically identical — unions are unordered and both property forms are valid JS. The `.d.ts` reordering is worth a second look regardless, since it rewrites the public type surface consumers compile against.

No tsconfig changes were needed. `rootDir` was deliberately **not** added to `tsconfig.json` — its `include` reaches `vite.config.ts` and `vitest.config.ts`, so it would raise `TS6059` for each; `tsconfig.build.json` already sets it. The `@typescript/typescript6` compat alias was skipped because nothing imports the compiler API once ESLint is gone.

### Outstanding — why this is a draft

1. **221 oxlint errors** after adopting the shared config, up from 151. The jump is mostly `categories: { correctness: 'error' }` and plugins the old config never enabled: `vitest/require-mock-type-parameters` (+24), `react/react-compiler` (+20), `no-console` (+7). They need triaging into follow-ups, or switching off locally with TODOs, before this can merge.
2. **`pre-commit-lint` runs `check-types`**, which now includes `lint` — so with those 221 errors outstanding, every commit is blocked. Either the errors get fixed first, or the hook stays on `check-tsc` and only CI runs the combined step. The four commits here used `--no-verify`.

Five of the new findings look real rather than noise: four redundant `x && x` self-guards (`Card.tsx:111`, `TextInput.tsx:124` and `:145`, `NumberInput.tsx:162`) and a missing React `key` in `Tag/storybook/Tag.stories.tsx:70`. ESLint never reported any of them.

One rule regressed: `@typescript-eslint/unbound-method` reported 2 errors on `ref={refs.setReference}` in `Tooltip.tsx`. `typescript/unbound-method` exists in oxlint and is accepted at `error`, but reports nothing there. oxlint flags the same two lines via `react/react-compiler` for a different reason, so the loss is judged acceptable.

- Skill and its findings: https://github.com/marshmallow-insurance/claude-code-plugins/pull/190
- Shared oxlint configs this consumes: https://github.com/marshmallow-insurance/campfire/pull/548

## Testing

No tests were added — this changes tooling, not source. Every existing check was baselined before any change and re-run after each phase:

| Check | Before | After |
|---|---|---|
| `check-tsc` | 0 errors (TS 5.9.3) | 0 errors (TS 7.0.2) |
| `vitest run` | 45 files, 246 tests pass | 45 files, 246 tests pass |
| `build` | 1191 files emitted | 1191 files, 123 differ cosmetically |
| lint | 215 problems (153 errors, 62 warnings) | 277 problems (221 errors, 56 warnings) |
| `@typescript-eslint/utils` | 4 copies in the tree | empty |

The lint baseline is from `npx eslint .`, not `npm run lint` — the script is `eslint . --fix`, so running it as written rewrites the tree and hides 151 formatting errors instead of reporting them. 18 files also failed `prettier --check` before any change here.

`oxlint-tsgolint@7.0.2001` was verified working under TypeScript 7.0.2 — type-aware output is identical before and after the compiler bump.


david-gomes5:
Superseded by a 2-PR stack that splits this by reviewable substance: #5119 (oxlint + campfire) then #5120 (TypeScript 5.9 → 7.0). Also fixes the lint script — `--format=checkstyle > file` left GitHub Actions with only an exit code, since nothing here consumes the XML; it now uses `--format=github` for inline annotations.

